PR #6, opened 2018-06-20
Add Notify gem to project

adrianclay:
We will need to send confirmation emails and SMSs back to the user
once they have signed up. We will be using GOV.UK Notify for this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'http://rubygems.org'
 ruby File.read('.ruby-version').chomp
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,6 +1,7 @@
 source 'http://rubygems.org'
 ruby File.read('.ruby-version').chomp
 
+gem 'notifications-ruby-client'
```
emileswarts: Should we lock down the versions for these gems?

PR #7, opened 2018-06-21
Generate user from self-signup email

RyanMacG:
When we receive a self signup email generate, store and return login details.


lib/user.rb
```diff
@@ -0,0 +1,41 @@
+class User < Sequel::Model(:userdetails)
+  User.unrestrict_primary_key
+  WORD_LIST = File.readlines(ENV['WORD_LIST_FILE']).map(&:strip)
+
+  def generate(email:)
```
adrianclay: Looks more like `find_or_create` than `generate`.

lib/user.rb
```diff
@@ -0,0 +1,41 @@
+class User < Sequel::Model(:userdetails)
+  User.unrestrict_primary_key
```
adrianclay: What's this all about?

lib/user.rb
```diff
@@ -0,0 +1,41 @@
...
+    return login_details(existing_user) if existing_user
+
+    user = create_user(email)
+    login_details(user)
+  end
+
+private
+
+  def random_username
+    username = ('a'..'z').to_a.sample(6).join
```
adrianclay: Wonder if we can remove this duplication on line 19.

lib/user.rb
RyanMacG: `find_or_create` is a method in both ActiveRecord and Sequel, I'd rather avoid any potential conflict/confusion by using it here

lib/user.rb
RyanMacG: because the username is the primary key for the table it can't be mass assigned in a `.create` unless we do this

lib/user.rb
RyanMacG: I could do

```ruby
def generate_username
  ('a'..'z').to_a.sample(6).join
end
```
but I don't think that saves us much

spec/lib/user_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+        expect(user_from_db.email).to eq(email)
+        expect(user_from_db.sponsor).to eq(email)
+      end
+    end
+  end
+
+  context 'avoiding duplicate usernames' do
+    before do
+      srand(2)
+      ('a'..'z').to_a.sample(6).join
```
adrianclay: What's this line do?

adrianclay: But the value isn't assigned anywhere which is what confuses me.

lib/user.rb
adrianclay: I think it looks a little nicer.

lib/user.rb
adrianclay: What do you mean by mass assigned?  Where do we do that in the code?

spec/lib/user_spec.rb
RyanMacG: Gotcha, removed it

lib/user.rb
RyanMacG: `User.create(username: random_username)...` is mass assignment

RyanMacG: https://github.com/jeremyevans/sequel/blob/master/doc/mass_assignment.rdoc

PR #8, opened 2018-06-21
Add check for the incoming email being authorised to sign up to GovWifi

adrianclay:
We currerntly only accept email self-signups from emails on a list of valid domain names.  The regex that gets used to check against gets is populated via ENV variable.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -14,5 +14,18 @@ class App < Sinatra::Base
     payload = JSON.parse request.body.read
 
     Net::HTTP.get(URI(payload['SubscribeURL'])) if payload['Type'] == 'SubscriptionConfirmation'
+    signup_user(JSON.parse(payload['Message'])) if payload['Type'] == 'Notification'
+    ""
+  end
+
+private
+
+  def signup_user(ses_notification)
+    from_address = ses_notification['commonHeaders']['from'][0]
+    User.new.generate(email: from_address) if authorised_email_domains_regex.match? from_address
+  end
+
+  def authorised_email_domains_regex
+    Regexp.new(ENV.fetch('AUTHORISED_EMAIL_DOMAINS_REGEX'))
   end
 end
--- a/spec/email_notification_spec.rb
+++ b/spec/email_notification_spec.rb
@@ -1,4 +1,6 @@
 RSpec.describe App do
+  before { ENV['AUTHORISED_EMAIL_DOMAINS_REGEX'] = "\.gov\.uk$" }
+
   describe 'POSTing a SubscriptionConfirmation to /user-signup/email-notification' do
     it 'makes a GET request to the SubscribeURL' do
       stub_request(:any, "www.example.com")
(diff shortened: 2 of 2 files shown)


spec/email_notification_spec.rb
```diff
@@ -9,21 +11,56 @@
...
+      it 'returns no sensitive information to SNS' do
+        post_notification
+        expect(last_response.body).to eq('')
+      end
+    end
+
+    describe 'from an unauthorised email address' do
+      let(:from_address) { 'adrian@madetech.com' }
+
+      it 'doesn\'t call CreateUser' do
```
RyanMacG: why not just use does not rather than escaping?

spec/email_notification_spec.rb
```diff
@@ -9,21 +11,56 @@
...
+
+    it 'returns a 200' do
+      post_notification
       expect(last_response).to be_ok
     end
+
+    describe 'from an authorised email address' do
+      let(:from_address) { 'adrian@adrian.gov.uk' }
+
+      it 'calls CreateUser' do
```
RyanMacG: shouldn't this be `generate`? Would expect this to match the method name

RyanMacG: `User#generate` if you want to include the class

spec/email_notification_spec.rb
```diff
@@ -9,21 +11,56 @@
   end
 
   describe 'POSTing a signup Notification to /user-signup/email-notification' do
-    it 'returns a 200' do
+    let(:from_address) { 'dummy@example.com' }
```
RyanMacG: should this succeed given it's not an authorised domain?

app.rb
```diff
@@ -14,5 +14,18 @@ class App < Sinatra::Base
...
     Net::HTTP.get(URI(payload['SubscribeURL'])) if payload['Type'] == 'SubscriptionConfirmation'
+    signup_user(JSON.parse(payload['Message'])) if payload['Type'] == 'Notification'
+    ""
+  end
+
+private
+
+  def signup_user(ses_notification)
+    from_address = ses_notification['commonHeaders']['from'][0]
+    User.new.generate(email: from_address) if authorised_email_domains_regex.match? from_address
```
RyanMacG: not a fan of this as a guard clause, I'd suggest `if authorised_email_domain?(from_address)` and do the regex match in there

RyanMacG:
Are you setting this as an ENV in the Dockerfile because I don't see it and would be good to have a default for development/test rather than having to set it everytime

RyanMacG:
You'll need to change the branch name to fix the failing jenkins

spec/email_notification_spec.rb
adrianclay: Succeed in what sense, just wanted to put in a dummy value for default.  The later tests set up specific values when asserting if `User#generate` has been called or not.

app.rb
adrianclay: Agreed.

adrianclay:
I don't like having a default in the ` Dockerfile ` as it could mean that the app is run without having the ENV overwritten by the infrastructure.  I've put in a default into the `docker-compose.yml`, as that would then give a default for local development.

I've also renamed the branch, see #9

PR #9, opened 2018-06-22
Add check for the incoming email being authorised to sign up to GovWifi

adrianclay:
We currerntly only accept email self-signups from emails on a list of valid domain names.  The regex that gets used to check against gets is populated via ENV variable.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -14,5 +14,22 @@ class App < Sinatra::Base
     payload = JSON.parse request.body.read
 
     Net::HTTP.get(URI(payload['SubscribeURL'])) if payload['Type'] == 'SubscriptionConfirmation'
+    signup_user(JSON.parse(payload['Message'])) if payload['Type'] == 'Notification'
+    ''
+  end
+
+private
+
+  def signup_user(ses_notification)
+    from_address = ses_notification['commonHeaders']['from'][0]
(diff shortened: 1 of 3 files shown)


app.rb
```diff
@@ -14,5 +14,22 @@ class App < Sinatra::Base
...
+
+private
+
+  def signup_user(ses_notification)
+    from_address = ses_notification['commonHeaders']['from'][0]
+    User.new.generate(email: from_address) if authorised_email_domain?(from_address)
+  end
+
+  def authorised_email_domain?(from_address)
+    authorised_email_domains_regex.match? from_address
```
RyanMacG: Can you put the argument in brackets here? Looks odd to have mixed styles

app.rb
```diff
@@ -14,5 +14,22 @@ class App < Sinatra::Base
     payload = JSON.parse request.body.read
 
     Net::HTTP.get(URI(payload['SubscribeURL'])) if payload['Type'] == 'SubscriptionConfirmation'
+    signup_user(JSON.parse(payload['Message'])) if payload['Type'] == 'Notification'
+    ""
```
RyanMacG: Any reason this is double quote rather than single? Odd to mix usage

PR #10, opened 2018-06-25
Pull in wordlist from S3 to overwrite the test wordlist

adrianclay:
We store the production wordlist in S3 to keep it secret.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -91,6 +91,15 @@ def deploy(deploy_environment) {
       sh('git fetch')
       sh('git checkout stable')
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -91,6 +91,15 @@ def deploy(deploy_environment) {
       sh('git fetch')
       sh('git checkout stable')
 
+      withAWS(credentials: 'jenkins-read-wordlist-credentials') {
+        s3Download(
+          file:'tmp/wordlist',
```
chrisblackburn: I note previously in this file spaces are used between named arguments and the values :-/

PR #11, opened 2018-06-26
Send the signup email via GOV.UK Notify

adrianclay:
We send an email to the user that signed up with details of their generated username and password so they can connect to GovWifi.

### DIFF
--- a/spec/email_notification_spec.rb
+++ b/spec/email_notification_spec.rb
@@ -41,11 +41,34 @@ RSpec.describe App do
     end
 
     describe 'from an authorised email address' do
+      let(:notify_email_url) { 'https://api.notifications.service.gov.uk/v2/notifications/email' }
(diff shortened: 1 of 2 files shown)


spec/email_notification_spec.rb
```diff
@@ -41,11 +41,34 @@ def post_notification
...
+      let(:username) { 'MockUsername' }
+      let(:password) { 'MockPassword' }
 
-      it 'calls User#generate' do
-        expect_any_instance_of(User).to receive(:generate).with(email: from_address)
+      before do
+        ENV['NOTIFY_API_KEY'] = notify_api_key
+        ENV['NOTIFY_USER_SIGNUP_EMAIL_TEMPLATE_ID'] = notify_template_id
+
+        expect_any_instance_of(User).to \
```
chrisblackburn: do you need \ in ruby? it's been a while

adrianclay: Because this statement is over several lines without the `\` ruby will assume that you are calling `.to` without any arguments.

PR #18, opened 2018-07-02
Rename the email parameter to contact

adrianclay:
The variable passed in will in future be either a phone number or email address, which we generalise as contact.


spec/lib/user_spec.rb
```diff
@@ -49,12 +49,12 @@
   context 'avoiding duplicate usernames' do
     before do
       srand(2)
-      User.new.generate(email: 'foo@bar.gov.uk')
+      User.new.generate(contact: 'foo@bar.gov.uk')
     end
 
     let!(:user) do
       srand(2)
```
emileswarts: What does this srand do?

adrianclay: [Seeds the system pseudo-random number generator, Random::DEFAULT, with number. The previous seed value is returned.](http://ruby-doc.org/core-2.0.0/Random.html#method-c-srand)

emileswarts: Feels like this User class has many responsibilities.
Would opt for having a password generator class on it's own, but can be a refactoring for the future.

lib/user.rb
```diff
@@ -30,12 +30,12 @@ def password_from_word_list
     WORD_LIST.sample(3).map(&:capitalize).join
   end
 
-  def create_user(email)
+  def create_user(contact)
```
emileswarts: This is referred to as a model in the tests, surprising to me to see a create_user method on this object.

PR #19, opened 2018-07-02
Add first slice for signing up via text message

adrianclay:
Calls the user generation method and sends details back to the user

Still outstanding is the "Help" flow.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -2,6 +2,7 @@ require 'sinatra/base'
 require 'net/http'
 
 require './lib/email_signup.rb'
+require './lib/sms_signup.rb'
 require './lib/user.rb'
 
 class App < Sinatra::Base
@@ -18,6 +19,11 @@ class App < Sinatra::Base
 
     Net::HTTP.get(URI(payload['SubscribeURL'])) if payload['Type'] == 'SubscriptionConfirmation'
     handle_signup_request(JSON.parse(payload['Message'])) if payload['Type'] == 'Notification'
+    ''
+  end
+
+  post '/user-signup/sms-notification' do
+    SmsSignup.new(user_model: User.new).execute(contact: params[:source])
     ''
   end
 
--- a/lib/sms_signup.rb
+++ b/lib/sms_signup.rb
@@ -1,0 +1,32 @@
+class SmsSignup
+  def initialize(user_model:)
+    @user_model = user_model
+  end
+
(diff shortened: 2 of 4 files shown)


app.rb
```diff
@@ -21,6 +22,11 @@ class App < Sinatra::Base
     ''
   end
 
+  post '/user-signup/sms-notification' do
+    SmsSignup.new(user_model: User.new).execute(contact: params[:source])
```
emileswarts: If this raised an exception would that be returned to FireText?

lib/sms_signup.rb
```diff
@@ -0,0 +1,32 @@
+class SmsSignup
+  def initialize(user_model:)
+    @user_model = user_model
+  end
+
+  def execute(contact:)
+    phone_number = normalised_phone_number(contact)
+    login_details = user_model.generate(contact: phone_number)
```
emileswarts: generate to me sounds like something stored in memory. find_or_create_by feels like a better name to me.  Probably a refactoring on its own

lib/sms_signup.rb
```diff
@@ -0,0 +1,32 @@
...
+    notify_params = { login: login_details[:username], pass: login_details[:password] }
```
emileswarts: Not sure this could ever be nil, but if it were, and nils some got into the database, it would be a successful authentication?  .fetch instead of [ would raise an error which might be better.

lib/sms_signup.rb
adrianclay: Yeah, that was my view.  See this comment here: https://github.com/alphagov/govwifi-user-signup-api/pull/7#discussion_r197150844

lib/sms_signup.rb
emileswarts: Oh sorry no, see this is for Notify.

app.rb
adrianclay: This is what Sinatra returns to the client a 500.

```
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/html;charset=utf-8
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< Content-Length: 30
< 
* Connection #0 to host localhost left intact
<h1>Internal Server Error</h1>% 
```

We'd get an entry inside of Sentry.

PR #20, opened 2018-07-03
Add template matcher for SMS messages

adrianclay:
We have a number of GOV.UK Notify templates for the SMS messages
depending on what message you send through to the service.

This behaviour of what you get sent has been matched to that of
the PHP app.

### DIFF
--- a/lib/sms_template_finder.rb
+++ b/lib/sms_template_finder.rb
@@ -1,0 +1,33 @@
+class SmsTemplateFinder
+  def execute(message_content:, env:)
+    device_name_matchers.each do |matcher, device_name|
+      return device_instruction_config(env).fetch(device_name) if message_content.match?(matcher)
+    end
+
+    template_name = 'generic_help'
+    template_name = 'credentials' if message_content.match?(/^\s*$/) || message_content.match(/go/i)
+    template_name = 'help_menu' if message_content.match?(/help/i)
+
+    config(env)[template_name]
+  end
+
+private
+
+  def device_instruction_config(env)
+    config(env).fetch('device_help')
+  end
+
+  def device_name_matchers
+    {
+      /1|android|samsung|galaxy|htc|huawei|sony|motorola|lg|nexus/i => 'android',
+      /2|ios|ipad|iphone|ipod/i => 'iphone',
+      /3|mac|OSX|apple/i => 'mac',
+      /4|win|windows/i => 'windows',
+      /5|blackberry/i => 'blackberry',
+    }
+  end
+
+  def config(env)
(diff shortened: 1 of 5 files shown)


lib/sms_template_finder.rb
```diff
@@ -0,0 +1,33 @@
+class SmsTemplateFinder
+  def execute(message_content:, env:)
+    device_name_matchers.each do |matcher, device_name|
+      return device_instruction_config(env).fetch(device_name) if message_content.match?(matcher)
+    end
+
+    template_name = 'generic_help'
+    template_name = 'credentials' if message_content == '' || message_content.match(/go/i)
+    template_name = 'help_menu' if message_content == 'help'
```
adrianclay: Do we want to make this a looser matcher?

lib/sms_template_finder.rb
```diff
@@ -0,0 +1,33 @@
+class SmsTemplateFinder
+  def execute(message_content:, env:)
+    device_name_matchers.each do |matcher, device_name|
+      return device_instruction_config(env).fetch(device_name) if message_content.match?(matcher)
+    end
+
+    template_name = 'generic_help'
+    template_name = 'credentials' if message_content == '' || message_content.match(/go/i)
```
adrianclay: Do we want to match if you send through a blank space?

lib/sms_template_finder.rb
```diff
@@ -0,0 +1,33 @@
...
+
+private
+
+  def device_instruction_config(env)
+    config(env).fetch('device_help')
+  end
+
+  def device_name_matchers
+    {
+      /1|android|samsung|galaxy|htc|huawei|sony|motorola|lg|nexus/i => 'android',
```
emileswarts: Would be nice to see all these regexes grouped and assigned to constants with descriptive names.

spec/lib/sms_template_finder_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+    end
+  end
+
+
+  context 'Given there is a Go message' do
+    it 'returns the user details template id' do
+      expect(template_for_message('go')).to eq('3a4b1ca8-7b26-4266-8b5f-e05fdbd11879')
+      expect(template_for_message('Go')).to eq('3a4b1ca8-7b26-4266-8b5f-e05fdbd11879')
+      expect(template_for_message('GO')).to eq('3a4b1ca8-7b26-4266-8b5f-e05fdbd11879')
```
emileswarts: Could refactor out this duplication with a let, ok with this for now.

PR #23, opened 2018-07-04
Log Incoming SMS Requests

emileswarts:
These will show up in CloudWatch for debugging if needed.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -24,6 +24,8 @@ class App < Sinatra::Base
   end
 
   post '/user-signup/sms-notification' do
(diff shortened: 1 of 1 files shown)


app.rb
```diff
@@ -24,6 +24,8 @@ class App < Sinatra::Base
   end
 
   post '/user-signup/sms-notification' do
+    logger.info("Processing SMS on /user-signup/sms-notification from #{params[:source]} with message #{params[:message]}")
```
adrianclay: It should be pretty clear that it is coming from that URL, but that's a minor point.

PR #26, opened 2018-07-05
Add SponsorUsers UseCase

adrianclay:
* Creates accounts for each user
  * Sends a confirmation email to the sponsor
  * Only allows government email addresses to sponsor new users

### DIFF
--- a/lib/sponsor_users.rb
+++ b/lib/sponsor_users.rb
@@ -1,0 +1,120 @@
+require_relative 'phone_number'
+
+class SponsorUsers
+  def initialize(user_model:)
+    @user_model = user_model
+  end
+
+  def execute(sponsees, sponsor)
+    validate_sponsees(sponsees)
+
+    sponsor_address = Mail::Address.new(sponsor).address
+
+    valid_sponsees = validate_sponsees(sponsees)
+    invite_sponsees(sponsor, sponsor_address, valid_sponsees)
+    send_confirmation_email(sponsor_address, valid_sponsees)
+  end
+
+private
+
+  def invite_sponsees(sponsor, sponsor_address, valid_sponsees)
+    valid_sponsees.each do |sponsee|
+      if valid_email?(sponsee)
+        sponsee_address = Mail::Address.new(sponsee).address
+        sponsor_email(sponsor, sponsor_address, sponsee_address)
+      else
+        sponsee_phone_number = PhoneNumber.internationalise_phone_number(sponsee)
+        sponsor_phone_number(sponsor_address, sponsee_phone_number)
+      end
+    end
+  end
+
+  attr_reader :user_model
+
+  def validate_sponsees(sponsees)
+    valid_sponsees = sponsees.select do |sponsee|
+      valid_email?(sponsee) || valid_phone_number?(sponsee)
+    end
+    valid_sponsees.uniq
+  end
+
+  def notify_client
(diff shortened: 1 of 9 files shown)


lib/sponsor_users.rb
```diff
@@ -0,0 +1,121 @@
...
+        sponsee_address = Mail::Address.new(sponsee).address
+        sponsor_email(sponsor, sponsor_address, sponsee_address)
+      else
+        sponsee_phone_number = PhoneNumber.internationalise_phone_number(sponsee)
+        sponsor_phone_number(sponsor_address, sponsee_phone_number)
+      end
+    end
+  end
+
+  attr_reader :user_model, :valid_sponsees
```
chrisblackburn: It look me a while to figure out what `valid_sponsees` was. Is it randomly hidden between two methods?

lib/sponsor_users.rb
```diff
@@ -0,0 +1,121 @@
...
+    validate_sponsees(sponsees)
+    invite_sponsees(sponsor, sponsor_address)
+    send_confirmation_email(sponsor_address)
+  end
+
+private
+
+  def invite_sponsees(sponsor, sponsor_address)
+    valid_sponsees.each do |sponsee|
+      if is_valid_email(sponsee)
```
chrisblackburn: is it more Ruby to use `valid_email?`?

adrianclay: Agreed.

lib/sponsor_users.rb
```diff
@@ -0,0 +1,121 @@
...
+        sponsee_phone_number = PhoneNumber.internationalise_phone_number(sponsee)
+        sponsor_phone_number(sponsor_address, sponsee_phone_number)
+      end
+    end
+  end
+
+  attr_reader :user_model, :valid_sponsees
+
+  def validate_sponsees(sponsees)
+    @valid_sponsees = sponsees.select do |sponsee|
```
chrisblackburn: As this has an `attr_reader` I don't think you need the @ prefix. Also I think when it's accessed earlier it doesn't so should probably pick one style

lib/sponsor_users.rb
```diff
@@ -0,0 +1,121 @@
...
+      is_valid_email(sponsee) || is_valid_phone_number(sponsee)
```
chrisblackburn: same question about the `valid_xx?` methods being Ruby naming convention

spec/lib/sponsor_users_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+    context 'On production' do
+      let(:environment) { 'production' }
+      let(:plural_sponsor_confirmation_template_id) { '58e8ef4a-ca6b-40cd-81df-ec9c781fed56' }
+
+      it 'Sends a multiple user confirmation email to the sponsor' do
+        expect(a_plural_sponsor_confirmation_request).to have_been_made.once
+      end
+    end
+
+    context 'On staging' do
```
chrisblackburn: Unclear of the purpose of this test. Is it double entry bookkeeping against the config values?

adrianclay: Yep, it's asserting picking up the environment correctly.

adrianclay:
I turned the `attr_reader` for `valid_sponsees` back into a parameter as it was causing confusionaroo.

PR #27, opened 2018-07-06
Add extractor for sponsees from an email message

adrianclay:
To sponsor guests for GovWifi you send an email with their details to sponsor@wifi.service.gov.uk

Each email/phone number appears on a new line.

The regression tests have been pulled over from the PHP test suite.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,9 @@
 source 'http://rubygems.org'
 ruby File.read('.ruby-version').chomp
 
-gem "mail", "~> 2.7"
+gem 'mail', '~> 2.7'
 gem 'mysql2'
+gem 'nokogiri', '~> 1.8'
 gem 'notifications-ruby-client', '~> 2.6.0'
 gem 'puma'
 gem 'sentry-raven'
--- a/lib/email_sponsees_extractor.rb
+++ b/lib/email_sponsees_extractor.rb
@@ -1,0 +1,31 @@
+require 'nokogiri'
+
+class EmailSponseesExtractor
+  def execute(email)
+    mail = Mail.read_from_string(email)
+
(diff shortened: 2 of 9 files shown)


lib/email_sponsees_extractor.rb
```diff
@@ -0,0 +1,31 @@
+require 'nokogiri'
+
+class EmailSponseesExtractor
+  def execute(s)
```
adrianclay: Ugg that variable name

lib/email_sponsees_extractor.rb
```diff
@@ -0,0 +1,31 @@
...
+      lines_from_html(mail)
+    end
+  end
+
+  def lines_from_plain_text(message)
+    message.decoded.lines "\n"
+  end
+
+  def lines_from_html(mail)
+    Nokogiri::HTML(mail.html_part.decoded).xpath("//text()").reject(&:cdata?).map do |node|
```
chrisblackburn: In the absence of alternative solutions, I'll trust that this is the cleanest way to achieve this :-)

adrianclay: I've moved the style tag check into the xpath, and it looks nicer I'd say.  Thoughts?

Gemfile
```diff
@@ -1,8 +1,9 @@
 source 'http://rubygems.org'
 ruby File.read('.ruby-version').chomp
 
-gem "mail", "~> 2.7"
+gem 'mail', "~> 2.7"
```
chrisblackburn: Don't forget the other double quotes

Gemfile
```diff
@@ -1,8 +1,9 @@
...
 gem 'mysql2'
+gem 'nokogiri', "~> 1.8"
```
chrisblackburn: Double quotes here also

PR #29, opened 2018-07-09
Hook up the SponsorUsers use case to the /email-notification/ route

adrianclay:
This integrates SponsorUsers up to the SES/SNS/S3 delivery mechanism.

We differentiate between signup and sponsor email requests, by the local part of the To address attached to the SES message, e.g.  `sponsor@wifi.service.gov.uk`

Also create a small S3 client which pulls down the raw email object from the bucket/object passed through from the SES notification details.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -2,6 +2,7 @@ require 'sinatra/base'
 require 'net/http'
 
 require './lib/email_signup.rb'
+require './lib/gateway/s3_object_fetcher.rb'
 require './lib/sms_response.rb'
 require './lib/sponsor_users.rb'
 require './lib/user.rb'
@@ -21,7 +22,7 @@ class App < Sinatra::Base
     payload = JSON.parse request.body.read
 
     Net::HTTP.get(URI(payload['SubscribeURL'])) if payload['Type'] == 'SubscriptionConfirmation'
-    handle_signup_request(JSON.parse(payload['Message'])) if payload['Type'] == 'Notification'
+    handle_notification(payload) if payload['Type'] == 'Notification'
     ''
   end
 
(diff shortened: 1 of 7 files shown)


app.rb
```diff
@@ -42,9 +43,35 @@ class App < Sinatra::Base
 
 private
 
+  def handle_notification(payload)
+    ses_notification = JSON.parse(payload['Message'])
+    if is_sponsor_request(ses_notification)
```
chrisblackburn: I thought we preferred the `sponsor_request?` style?

app.rb
```diff
@@ -42,9 +43,35 @@ class App < Sinatra::Base
 
 private
 
+  def handle_notification(payload)
+    ses_notification = JSON.parse(payload['Message'])
```
chrisblackburn: Is this the place that knows about implementation specifics such as SES?

spec/lib/email_sponsees_extractor_spec.rb
```diff
@@ -50,27 +53,28 @@
...
 
     it 'Multiple levels of multipart messages' do
-      sponsees = test_case 'email-sponsor-multilevel-multipart'
+      test_case 'email-sponsor-multilevel-multipart'
       expect(sponsees.first).to eq('example.user2@example.co.uk')
     end
 
     it 'Base64 encoded message' do
-      sponsees = test_case('email-sponsor-base64')
+      test_case('email-sponsor-base64')
```
chrisblackburn: I find the inconsistent use parentheses mildly upsetting

app.rb
adrianclay: Yes, currently all knowledge of SES remains inside of `app.rb`.

spec/lib/email_sponsees_extractor_spec.rb
adrianclay: Yeah, that is odd.

PR #33, opened 2018-07-10
Update README with up to date information

adrianclay:
Now that the journeys have been implemented, add some background around what this repository does and info on how to deploy changes.


emileswarts:
Would be good to document the help routes as well.  So they don't get lost in the abyss like before.

PR #36, opened 2018-07-12
Introduce a Contact Sanitiser

emileswarts:
Restore original email sanitiser regex to ensure that emails are in a
consistent format for sending signup invites to sponsees.

Previously the system was rejecting emails in the format of
```
clay@example.com<mailto:clay@example.com>
```
Office365 was formatting email addresses in plain text form like
this.

This class will ensure emails are in a valid format along with phone
numbers.  It will also internationalise phone numbers.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -2,6 +2,7 @@ require 'sinatra/base'
 require 'net/http'
 
 require './lib/email_signup.rb'
+require './lib/contact_sanitiser.rb'
 require './lib/gateway/s3_object_fetcher.rb'
 require './lib/sms_response.rb'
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -18,3 +18,5 @@ services:
       DB_HOSTNAME: db
       RACK_ENV: development
       AUTHORISED_EMAIL_DOMAINS_REGEX: "\\.gov\\.uk$$"
+    volumes:
```
adrianclay: Feels odd introducing this in the same commit.

PR #45, opened 2018-07-16
Use better structure and require files automatically

naliwajek:
* Use better structure for the codebase: /use_cases, /gateways and /model dirs
* Require all files automatically thank to require_all gem


app.rb
```diff
@@ -1,15 +1,8 @@
...
-require './lib/sns_notification_handler.rb'
-require './lib/email_signup.rb'
-require './lib/contact_sanitiser.rb'
-require './lib/gateway/s3_object_fetcher.rb'
-require './lib/sms_response.rb'
-require './lib/sponsor_users.rb'
-require './lib/user.rb'
-require './lib/sms_template_finder.rb'
-require './lib/email_sponsees_extractor.rb'
+require_all 'lib'
```
emileswarts: nice

PR #46, opened 2018-07-16
Add gateway for user signups statistics

naliwajek:
It allows to fetch signups for today and total number of
signups before end of today.

### DIFF
--- a/spec/lib/gateway/stat_gateway_spec.rb
+++ b/spec/lib/gateway/stat_gateway_spec.rb
@@ -5,7 +5,11 @@ describe StatGateway do
 
   context 'given no signups' do
     it 'returns zero signups for today' do
-      expect(subject.signups).to eq(today: 0, total: 0)
+      expect(subject.signups[:today]).to eq(0)
+    end
+
+    it 'returns zero signups for today' do
+      expect(subject.signups[:total]).to eq(0)
     end
   end
 
@@ -16,8 +20,12 @@ describe StatGateway do
       end
     end
 
-    it 'returns zero signups for today' do
-      expect(subject.signups).to eq(today: 3, total: 3)
+    it 'returns signups for today' do
+      expect(subject.signups[:today]).to eq(3)
+    end
+
+    it 'returns same total number of signups' do
+      expect(subject.signups[:total]).to eq(3)
     end
   end
 
@@ -30,8 +38,12 @@ describe StatGateway do
       end
     end
 
-    it 'returns zero signups 5 signups for today and 6 total' do
-      expect(subject.signups).to eq(today: 5, total: 6)
+    it 'returns zero signups 5 signups for today' do
+      expect(subject.signups[:today]).to eq(5)
+    end
+
+    it 'returns zero signups 6 signups total' do
+      expect(subject.signups[:total]).to eq(6)
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,49 @@
+describe StatGateway do
+  before do
+    DB[:userdetails].truncate
```
adrianclay: Be nice to remove this duplication of truncate at some point, maybe using a truncation gem?

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    it 'returns zero signups for today' do
+      expect(subject.signups).to eq(today: 5, total: 6)
+    end
+  end
+
+  context 'given 2 signups tomorrow' do
+    before do
+      2.times do |i|
+        User.create(username: "Today #{i}", created_at: Date.today + 1)
```
adrianclay: "Today #{i}" is a bit misleading for this testcase.

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  context 'given 5 signups today and 1 yesterday' do
+    before do
+      User.create(username: 'Yesterday', created_at: Date.today - 1)
+
+      5.times do |i|
+        User.create(username: "Today #{i}")
+      end
+    end
+
+    it 'returns zero signups for today' do
```
adrianclay: ` it 'returns 5 signups for today and 6 in total' ` ?

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  context 'given 3 signups today' do
+    before do
+      3.times do |i|
+        User.create(username: "Today #{i}")
+      end
+    end
+
+    it 'returns zero signups for today' do
```
adrianclay: ` it 'returns 3 signups for today and 3 in total' ` ?

adrianclay:
I'd add a bit of spiel to the commit message about using these stats and pushing them into the "performance platform".

naliwajek:
@adrianclay none of those commits are doing any pushing, though.

What if this gateway is going to be used in the future for pushing stats to other systems? Are we going to go back to files and churn them to update theirs, no longer valid, commit message?

Separation of concerns on code level should correspond to commit messages as well IMO. If using the gateway for a new system doesn't force us to update it (that's the whole point of correct flow of control) then I don't see why we should include a message about upstream components' feature.

Use Case commit message will describe this and that should be a starting point for anyone who wants to know what app is doing.

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    it 'counds them against SMS total signups' do
+      expect(subject.signups[:sms_total]).to eq(2)
+    end
+
+    it 'counds them against SMS today\'s signups' do
+      expect(subject.signups[:sms_today]).to eq(1)
+    end
+  end
+end
```
naliwajek: Some of those tests will quickly become redundant and noisy once we add two last stats. I don't want to create a mega PR so I ought to merge it now and refactor once we have the whole gateway working.

spec/lib/gateway/stat_gateway_spec.rb
naliwajek: Agree.

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+    it 'counds them against SMS total signups' do
+      expect(subject.signups[:sms_total]).to eq(1)
+    end
+
+    it 'counds them against SMS today\'s signups' do
+      expect(subject.signups[:sms_today]).to eq(1)
+    end
+  end
+
+  context 'given SNS signups made on different dates' do
```
adrianclay: Typo, in SNS.

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+        contact: '+0123456789',
+        sponsor: '+0123456789'
+        )
+    end
+
+    it 'counts them against total singups' do
+      expect(subject.signups[:total]).to eq(2)
+    end
+
+    it 'counds them against today\'s signups' do
```
adrianclay: Type on counds.

adrianclay: Also don't think you need the apostrophe, here, but don't trust my English.

adrianclay:
I'd argue that context is super important, and that's what people reading commit messages from the past will find useful.

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+    it 'returns zero signups for today' do
+      expect(subject.signups[:today]).to eq(0)
+    end
+
+    it 'returns zero signups total' do
+      expect(subject.signups[:today]).to eq(0)
+    end
+  end
+
+  context 'given 1 SMS signup today' do
```
adrianclay: Given 1 sms signup and 1 email signup?

I'd then also be explicit around setting the contact field.

spec/lib/gateway/stat_gateway_spec.rb
adrianclay: Sounds like a plan.

PR #47, opened 2018-07-17
Add email and sponsored signups to statistics

naliwajek:
* add email signups to statistics
* add sponsored signups to statistics
* introduce `SignUp` model to deal with Sequel queries for statistics
* and use that model as a repository for a gateway (not injected because it's fairly simple so far)

### DIFF
--- a/lib/gateway/stat_gateway.rb
+++ b/lib/gateway/stat_gateway.rb
@@ -12,37 +12,31 @@ class StatGateway
 
 private
 
+  def repository
+    SignUp
+  end
+
   def signups_today
-    User
-      .where(Sequel[:created_at] > Date.today)
-      .where(Sequel[:created_at] < Date.today + 1)
+    repository.all.today
   end
 
   def signups_total
-    User.where(Sequel[:created_at] < Date.today + 1)
+    repository.all
   end
 
   def sms_signups_today
-    signups_today
-      .where(Sequel.like(:contact, '+%'))
-      .where(contact: Sequel[:sponsor])
+    signups_today.self_sign.with_sms
   end
 
   def sms_signups_total
-    signups_total
-      .where(Sequel.like(:contact, '+%'))
-      .where(contact: Sequel[:sponsor])
+    signups_total.self_sign.with_sms
   end
 
   def email_signups_today
-    signups_today
-      .where(Sequel.like(:contact, '%@%'))
-      .where(contact: Sequel[:sponsor])
+    signups_today.self_sign.with_email
   end
 
   def email_signups_total
-    signups_total
-      .where(Sequel.like(:contact, '%@%'))
-      .where(contact: Sequel[:sponsor])
+    signups_total.self_sign.with_email
   end
 end
--- a/lib/gateway/stat_gateway.rb
+++ b/lib/gateway/stat_gateway.rb
@@ -2,11 +2,13 @@ class StatGateway
   def signups
     {
       today: signups_today.count,
-      total: signups_total.count,
+      cumulative: signups_cumulative.count,
       sms_today: sms_signups_today.count,
-      sms_total: sms_signups_total.count,
+      sms_cumulative: sms_signups_cumulative.count,
       email_today: email_signups_today.count,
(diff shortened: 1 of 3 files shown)


lib/gateway/stat_gateway.rb
```diff
@@ -4,31 +4,39 @@ def signups
       today: signups_today.count,
       total: signups_total.count,
       sms_today: sms_signups_today.count,
-      sms_total: sms_signups_total.count
+      sms_total: sms_signups_total.count,
+      email_today: email_signups_today.count,
+      email_total: email_signups_total.count
     }
```
naliwajek: Side note: I like word `total` on a hash key. Makes sense because that's a number. I dislike it as a suffix for a method name... because that method returns a query, not a *total*-of-anything value. May change it to `all` later...

naliwajek:
Performance-wise in tests, it'd make sense to create all signups at the beginning instead of truncating and re-creating them in each context. I'll make a refactor once all stats are in.

lib/gateway/stat_gateway.rb
adrianclay: Why not make the change now, i'd argue that `cumulative_email_signups` is more readable.

lib/model/sign_up.rb
```diff
@@ -0,0 +1,23 @@
+class SignUp < Sequel::Model(:userdetails)
+  dataset_module do
```
adrianclay: What does ` dataset_module ` add?  Not sure I understand it.  I like this refactoring though.

spec/lib/gateway/stat_gateway_spec.rb
```diff
@@ -123,4 +141,37 @@
...
+
+    it 'counts them against todays signups' do
+      expect(subject.signups[:today]).to eq(1)
+    end
+
+    it 'counts them against email total signups' do
+      expect(subject.signups[:email_total]).to eq(2)
+    end
+
+    it 'counts them against email todays signups' do
```
adrianclay: This feels a little jarring.

`counts them against todays email signups` is nicer.

lib/model/sign_up.rb
```diff
@@ -0,0 +1,23 @@
...
+    def all
```
emileswarts: not sure .all describes exactly what is going on here.
.last_two_days would be more descriptive IMO

lib/model/sign_up.rb
naliwajek: I found a nice cheat sheet here:

https://gist.github.com/odigity/2d8123a90e8c2331d0ad#file-sequel_scopes-rb-L5

lib/model/sign_up.rb
naliwajek: That's everything earlier than tomorrow – a way to ensure we don't use corrupted data from the future.

So, yeah, it ain't `all`, you're right. But it's not `.last_two_days` either.

lib/gateway/stat_gateway.rb
naliwajek: Where's the accumulation?

lib/model/sign_up.rb
naliwajek: `*everything earlier than tomorrow* == `.current` ?

naliwajek: Or not... `current` sounds like *logged in right now* 👎

lib/gateway/stat_gateway.rb
```diff
@@ -2,33 +2,51 @@ class StatGateway
...
-      total: signups_total.count,
+      cumulative: signups_cumulative.count,
       sms_today: sms_signups_today.count,
-      sms_total: sms_signups_total.count
+      sms_cumulative: sms_signups_cumulative.count,
+      email_today: email_signups_today.count,
+      email_cumulative: email_signups_cumulative.count,
+      sponsored_today: sponsored_signups_today.count,
+      sponsored_cumulative: sponsored_signups_cumulative.count
     }
```
naliwajek: Is that use of word `cumulative` okay? @adrianclay

PR #48, opened 2018-07-17
Send stats to Performance service

naliwajek:


app.rb
```diff
@@ -2,6 +2,8 @@
 require 'net/http'
 require 'require_all'
 
+module Common; end
```
naliwajek: Will introduce a proper `namespaces.rb` in a PR with structure refactor 👍

PR #49, opened 2018-07-18
Send volumetrics stats to Performance

naliwajek:


lib/presenter/performance_platform.rb
```diff
@@ -0,0 +1,47 @@
...
+      count: count,
+      cumulative_count: cumulative_count
+    }
+  end
+
+  def encode_id(channel)
+    Common::Base64.encode_array(
+      [
+        timestamp,
+        'govwifi',
```
naliwajek: Need to verify is this `govwifi` or `gov-wifi` 🤔

lib/gateway/performance_platform_gateway.rb
```diff
@@ -0,0 +1,20 @@
...
+    uri = URI("#{ENV['PERFORMANCE_URL']}data/gov-wifi/volumetrics")
+
+    post(uri, data)
+  end
+
+private
+
+  def post(uri, data)
+    request = Net::HTTP::Post.new(uri)
+    request['Authorization'] = "Bearer #{ENV['PERFORMANCE_BEARER_VOLUMETRICS']}"
```
adrianclay: This gateway is currently coupled to volumetrics.

naliwajek: Yeah, I don't want to go too far with over-architecting this now. When we add different stats, only then we will think about how to DI them to the gateway.

PR #50, opened 2018-07-18
Add rake task which calls publish statistics to the perf platform

adrianclay:
As part of integrating Rake into the project, we've pulled out some
common bootstrap code into loader.rb which removes duplication
between Rack, Rake amd RSpec.

This rake task will be called by a scheduled task run every night.


lib/loader.rb
```diff
@@ -0,0 +1,15 @@
...
+
+DB = Sequel.connect(
+  adapter: 'mysql2',
+  host: ENV.fetch('DB_HOSTNAME'),
+  database: ENV.fetch('DB_NAME'),
+  user: ENV.fetch('DB_USER'),
+  password: ENV.fetch('DB_PASS')
+)
+
+module Common; end
```
naliwajek: Let's extract `namespaces.rb` once we follow `lib/<actor>` pattern 🎉

PR #53, opened 2018-07-19
[WIP] Add completion rate statistics

naliwajek:


lib/performance_platform/presenter/report.rb
```diff
@@ -5,12 +5,15 @@ def initialize(stats:)
...
-    ]
+    {
+      metric_name: stats[:metric_name],
+      payload: [
+        as_hash(stats[:today], stats[:cumulative], 'all-sign-ups'),
+        as_hash(stats[:sms_today], stats[:sms_cumulative], 'sms-sign-ups'),
+        as_hash(stats[:email_today], stats[:email_cumulative], 'email-sign-ups'),
+        as_hash(stats[:sponsored_today], stats[:sponsored_cumulative], 'sponsored-sign-ups'),
+      ]
+    }
```
naliwajek: `as_hash` needs to be extract out of here as it will differ per gateway.

lib/performance_platform/gateway/completion_rate.rb
```diff
@@ -0,0 +1,44 @@
+class PerformancePlatform::Gateway::CompletionRate
+  def fetch_stats
+    {
+      period: 'week',
```
adrianclay: Feels a little counterintuitive that this is returned and not passed in as an argument, but that's a minor point.

I'd expect the users of this gateway to be deciding over what period they want this statistic (not the other way round).

lib/performance_platform/gateway/completion_rate.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+private
+
+  def repository
+    PerformancePlatform::Repository::SignUp
+  end
+
+  def sms_registered
+    repository.with_sms.last_week
```
adrianclay: Love these 1 line methods

lib/performance_platform/gateway/completion_rate.rb
```diff
@@ -0,0 +1,44 @@
...
+  def sms_logged_in
+    sms_registered.with_sessions
+  end
+
+  def email_logged_in
+    email_registered.with_sessions
+  end
+
+  def sponsor_logged_in
+    sponsor_registered.with_sessions
```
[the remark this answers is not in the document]
adrianclay: Maybe `.logged_in` ?

lib/performance_platform/gateway/completion_rate.rb
naliwajek: I'd agree but since so far we don't have a need for it...

lib/performance_platform/repository/sign_up.rb
```diff
@@ -23,5 +23,13 @@ def with_sms
     def with_email
       where(Sequel.like(:contact, '%@%'))
     end
+
+    def last_week
+      where(created_at: (Date.today - 14)..(Date.today - 7))
+    end
+
+    def with_sessions
+      where(Sequel.lit('last_login IS NOT NULL'))
```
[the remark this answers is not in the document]
naliwajek: Huh, I saw that `NOT NULL` example in Sequel docs 😱 

Good catch, will change.

PR #55, opened 2018-07-23
Add completionRate stats

naliwajek:


lib/performance_platform/gateway/performance_report.rb
```diff
@@ -24,4 +24,14 @@ def post(uri, data)
       JSON.parse(response.body)
     end
   end
+
+  def build_url(data)
+    URI("#{ENV['PERFORMANCE_URL']}data/gov-wifi/#{data[:metric_name]}")
```
emileswarts: I always opt for .fetch over [
Just to avoid a nil getting in there if the ENV isn't set correctly

lib/performance_platform/gateway/completion_rate.rb
```diff
@@ -0,0 +1,44 @@
...
+  def fetch_stats
+    {
+      period: 'week',
+      metric_name: 'completion-rate',
+      sms_registered: sms_registered.count,
+      sms_logged_in: sms_logged_in.count,
+      email_registered: email_registered.count,
+      email_logged_in: email_logged_in.count,
+      sponsor_registered: sponsor_registered.count,
+      sponsor_logged_in: sponsor_logged_in.count,
```
emileswarts: Loving this DSL here

lib/performance_platform/gateway/performance_report.rb
naliwajek: Do we want this *not* to blow up when ENV variables are wrong?

naliwajek: (I mean, I can see both approaches as legitimate strategies.)

lib/performance_platform/presenter/completion_rate.rb
```diff
@@ -0,0 +1,52 @@
...
+      [
+        timestamp,
+        'gov-wifi',
+        stats[:period],
+        stats[:metric_name],
+        stage,
+        channel
+      ]
+    )
+  end
```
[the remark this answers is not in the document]
naliwajek: But not right now.

spec/unit/lib/performance_platform/gateway/completion_rate_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      username: '8',
+      created_at: Date.today - 10,
+      contact: 'me@example.com',
+      sponsor: 'me@example.com',
+      last_login: Date.today
+    )
+  end
+
+  context 'given no signups' do
+    it 'returns stats with zero signups' do
```
emileswarts: Are these stats below for 0 signups?
Are sms_registered and email_registered signups?

naliwajek: Mhm, no, just a copy left over, good catch 👍

PR #56, opened 2018-07-23
Extract Performance dataset to ENV variable

naliwajek:
We send performance stats to many different datasets inside Performance service.


emileswarts:
LGTM

PR #61, opened 2018-07-26
Send signup statistics from yesterday not today

emileswarts:
Based on the current functionality, we only send statistics from the day
before, not today.

Fix date comparison functionality to compare against date only, not
time.


kjdchapman:
LGTM!

PR #62, opened 2018-07-27
Use ENV variable for Performance dataset

naliwajek:


kjdchapman:
Why is this good? 

What does this let us do in the future?

naliwajek:
Change to what dataset we send data with terraform config instead of churning code files.

naliwajek:
It would work previously as well but now encoded IDs will be consistent with a dataset in which they're saved. Just to keep our minds sane.

PR #88, opened 2018-10-01
[WIP]: Change reply address from support to do-not-reply

Renny4Real:
- Checking notify documentation has given a hint in the right direction in implementing a change of reply address when the end user clicks reply to a sign-up notification email. This would involved adding an additional address and using the generated token in the email sign up use-case.

- In pursuit of this, attempts were made to create a suitable test. The thought process was to create a test that returned or demonstrated that a particular field in the email template e.g. email address, was present in the request. From here, it could be determined how to implement the change in the reply addresses in a more confident manner. 

- Suggestions would be [greatly] appreciated for methods to test an instance of the user trying to create a reply email and checking that the reply email given is the desired one.

### DIFF
--- a/lib/wifi_user/use_case/email_signup.rb
+++ b/lib/wifi_user/use_case/email_signup.rb
@@ -26,10 +26,15 @@ private
       email_address: email_address,
       template_id: credentials_template_id,
       personalisation: login_details
+      # email_reply_to_id: do_not_reply_template_id
     )
   end
 
   def credentials_template_id
     YAML.load_file("config/#{ENV['RACK_ENV']}.yml").fetch('notify_email_template_ids').fetch('self_signup_credentials')
   end
+
+  # def do_not_reply_template_id
+  #   YAML.load_file("config/#{ENV['RACK_ENV']}.yml").fetch('notify_email_template_ids').fetch('do_not_reply_credentials')
+  # end
 end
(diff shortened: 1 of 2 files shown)


lib/wifi_user/use_case/email_signup.rb
```diff
@@ -25,11 +25,17 @@ def send_signup_instructions(email_address, login_details)
...
+      personalisation: login_details,
+      email_reply_to_id: do_not_reply_template_id
     )
   end
 
   def credentials_template_id
     YAML.load_file("config/#{ENV['RACK_ENV']}.yml").fetch('notify_email_template_ids').fetch('self_signup_credentials')
   end
+
+  def do_not_reply_template_id
```
tommotaylor: I think `do_not_reply_email_address_id` (or similar) would be a better name. The `template_id` is the ID of the actual email that Notify sends.

config/staging.yml
```diff
@@ -11,6 +11,7 @@ notify_sms_template_ids:
 
 notify_email_template_ids:
   self_signup_credentials: 96d1f5ac-2ebe-41a7-878f-9a569e0bb55c
+  do_not_reply_credentials: 45d6b6c4-6a36-47df-b34d-256b8c0d1511
```
tommotaylor: I don't think the name is quite right. Or perhaps it shouldn't be nested. The value above (`self_signup_credentials`) is a template (https://www.notifications.service.gov.uk/services/8a09d848-d453-4aea-bb1e-46f7d19a1814/templates/96d1f5ac-2ebe-41a7-878f-9a569e0bb55c) whereas the new `do_not_reply_credentials` is just an ID for an email address. 

So I don't think it should be nested under `notify_email_template_ids:` because its not a template ID.

tommotaylor:
@georgeschena @Renny4Real Two questions:

1) How does this effect all our other Notify mailers? Do we need to set the `email_reply_to_id` on every request to Notify or will it default to the `govwifi-support@` email address?

2) The naming of the new key/value isn't quite right, or its in the wrong place. I left a comment. It currently looks like a template ID but its not.

PR #92, opened 2018-10-11
Add test coverage reporting

cybojenix:


booyaa:
Adding for reference, our reasoning behind using coverage tools: https://github.com/alphagov/govwifi-admin/pull/189#issuecomment-428997078

PR #96, opened 2018-10-23
Set log level based on environment

booyaa:
**What**

Explicitly set the logging level for Production to `INFO`; Development and Staging to `DEBUG`.

**Why**

Required for [Live Services Supportability][link_lss].

[link_lss]: https://github.com/madetech/live-services-field-guide/blob/master/processes/checklists/supportability.md


app.rb
```diff
@@ -6,6 +6,10 @@
 class App < Sinatra::Base
   configure :production, :staging, :development do
     enable :logging
+
```
cybojenix:
```suggestion
configure do
  enable :logging
  set :logging, Logger::DEBUG
end
configure :production do
  set :logging, Logger::INFO
end
```

cybojenix: seems suggested changes don't work on multi-lines yet. Beware .

PR #97, opened 2018-10-23
Ensures app always waits for db

booyaa:
**What**

Uses `wait_for_mysql` to delay starting up the `app` service when running `make serve`

**Why**

`app` frequently start up before `db` is ready causing it to exit early.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,9 @@ build:
 
 serve:
 	$(MAKE) build
-	docker-compose up -d
+	docker-compose up -d db
+	./mysql/bin/wait_for_mysql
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,7 +13,9 @@ build:
 
 serve:
 	$(MAKE) build
-	docker-compose up -d
+	docker-compose up -d db
+	./mysql/bin/wait_for_mysql
+	docker-compose up
```
kjdchapman: Looks like the jenkinsfile is using this command at the moment, and the lack of a `-d` flag is causing it to hang.

PR #124, opened 2019-01-02
User deletion

emileswarts:
This GDPR cleanup used to be done by a lambda.

We've since decided that all cron tasks will be run as ECS scheduled
tasks.  This is the rewrite of it from Python to Ruby.


lib/gdpr/use_case/delete_users.rb
```diff
@@ -0,0 +1,11 @@
+class Gdpr::UseCase::DeleteUsers
```
kjdchapman: I'd call this `DeleteStaleUsers` or `DeleteOldUsers` - something like that?

.rubocop.yml
```diff
@@ -0,0 +1,3 @@
+Lint/AmbiguousBlockAssociation:
```
kjdchapman: I feel like I ran into this the other day and managed to fix it. I'll have a quick play to see if I can?

lib/gdpr/use_case/obfusticate_sponsors.rb
```diff
@@ -0,0 +1,11 @@
+class Gdpr::UseCase::ObfusticateSponsors
```
kjdchapman: spelling, `Obfusticate` -> `Obfuscate`. This is in many other places too, but I think it's all the same misspelling, so find and replace should hopefully be e-z.

lib/gdpr/gateway/usedetails.rb
```diff
@@ -0,0 +1,15 @@
...
+    DB.run('DELETE FROM userdetails
+        WHERE (last_login < DATE_SUB(NOW(), INTERVAL 12 MONTH)
+        OR (last_login IS NULL AND created_at < DATE_SUB(NOW(), INTERVAL 12 MONTH)))
+        AND username != "HEALTH"')
+  end
+
+  def obfusticate_sponsors
+    DB.run("UPDATE userdetails ud1
+        LEFT JOIN userdetails as ud2 ON ud1.sponsor = ud2.contact
+        SET ud1.sponsor = REPLACE(ud1.sponsor, SUBSTRING_INDEX(ud1.sponsor, '@', '1'), 'user')
```
kjdchapman: Just to double-check my understanding of the SQL black magic - this replaces `kyle@arbitrary-domain.com` with `user@arbitrary-domain.com`?

.rubocop.yml
kjdchapman: Ah, I see the issue you referenced in commit message. Alright, this seems fine to me.

lib/gdpr/gateway/usedetails.rb
emileswarts: Yup

lib/gdpr/use_case/delete_users.rb
emileswarts: Fair enough, called it DeleteInactiveUsers

PR #125, opened 2019-01-02
Rename delete_users to delete_inactive_users

emileswarts:
It is not obvious, from the name that this is a safe operation and
only removes users to keep use GDPR compliant


README.md
```diff
@@ -45,6 +45,18 @@ aws ecs run-task --cluster api-cluster --task-definition user-signup-api-task --
 aws ecs run-task --cluster api-cluster --task-definition user-signup-api-task --count 1 --overrides "{ \"containerOverrides\": [{ \"name\": \"user-signup\", \"command\": [\"bundle\", \"exec\", \"rake\", \"publish_weekly_statistics['2018-05-03']\"] }] }" --region eu-west-2
 ```
 
+## GDPR
+
+#### Inactive User Deletion
+
+Any user who has not logged into GovWifi for more than 12 months is considered inactive.
+
+We have a rake tasks that is running daily to ensure this happens
```
kjdchapman: nit: typo
```suggestion
We have a rake task that runs daily to ensure this happens
```

PR #130, opened 2019-01-08
Use sensible_logging gem

booyaa:
Sets up [sensible_logging](https://github.com/madetech/sensible_logging) gem which provides:

* Tagged logging
* Request ID (for tagged logging, and to include in Sentry exceptions in future)
* Minimal request log format

Example output from request logging:

```
[localhost] [development] [76b706bf-7f0c-48ec-8045-57a40d5aad82] method=GET path=/healthcheck status=200 duration=0.0207032 params={"test"=>"one", "two"=>"three"}
```

If you reference the built in Sinatra `logger` helper within the app/routes, then you'll see output like:

```
get 'healthcheck' do
  logger.debug('testing')
  'Healthcheck'
end
```

```
[localhost] [development] [76b706bf-7f0c-48ec-8045-57a40d5aad82] testing
[localhost] [development] [76b706bf-7f0c-48ec-8045-57a40d5aad82] method=GET path=/healthcheck status=200 duration=0.0207032 params={"test"=>"one", "two"=>"three"}
```


config.ru
```diff
@@ -1,2 +1,7 @@
+#\ --quiet
```
kjdchapman: Is this functionality that could be folded into the gem at some point? Or are we aiming to keep the gem server-agnostic?

booyaa: I've raised it as an issue on sensible_logging to see if we can override rack: https://github.com/madetech/sensible_logging/issues/24

If we can we'll update the code in these APIs accordingly.

PR #138, opened 2019-02-05
Write multiple dbs

emileswarts:


kjdchapman:
If I'm reading this right, some of the writing to these dbs is non-deterministic - e.g. generating passwords for users.

Will it be a problem that they're out of sync when we switch back?

PR #146, opened 2019-02-26
Use new user signup whitelist

emileswarts:
We made the user signup white list content manageable in the admin.
It publishes a regex to S3.  We can use this regex instead of the
environment variable set previously.

### DIFF
--- a/lib/wifi_user/use_case/check_if_whitelisted_email.rb
+++ b/lib/wifi_user/use_case/check_if_whitelisted_email.rb
@@ -1,0 +1,21 @@
+module WifiUser
+  module UseCases
+    class CheckIfWhitelistedEmail
+      def initialize(gateway:)
+        @email_regex_gateway = gateway
+      end
+
+      def execute(email)
+        result = email_regex_gateway.fetch
+
+        pattern = Regexp.new(result, Regexp::IGNORECASE)
+
+        { success: email.to_s.match?(pattern) }
+      end
+
+    private
+
+      attr_reader :email_regex_gateway
+    end
+  end
+end
--- a/lib/wifi_user/use_case/email_signup.rb
+++ b/lib/wifi_user/use_case/email_signup.rb
@@ -2,14 +2,15 @@ require 'mail'
 require 'notifications/client'
 
 class WifiUser::UseCase::EmailSignup
-  def initialize(user_model:)
+  def initialize(user_model:, s3_gateway:)
     @user_model = user_model
+    @s3_gateway = s3_gateway
   end
 
   def execute(contact:)
     email_address = Mail::Address.new(contact).address
 
-    return unless Common::EmailAddress.authorised_email_domain?(email_address)
+    return unless authorised_email_domain?(email_address)
 
     login_details = user_model.generate(contact: email_address)
     send_signup_instructions(email_address, login_details)
@@ -17,7 +18,12 @@ class WifiUser::UseCase::EmailSignup
 
 private
 
-  attr_accessor :user_model
+  attr_accessor :user_model, :s3_gateway
+
+  def authorised_email_domain?(email)
+    regex = Regexp.new(s3_gateway.fetch, Regexp::IGNORECASE)
+    regex.match?(email)
+  end
 
   def send_signup_instructions(email_address, login_details)
     client = Notifications::Client.new(ENV.fetch('NOTIFY_API_KEY'))
--- a/lib/wifi_user/use_case/email_signup.rb
+++ b/lib/wifi_user/use_case/email_signup.rb
@@ -2,36 +2,33 @@ require 'mail'
 require 'notifications/client'
 
 class WifiUser::UseCase::EmailSignup
-  def initialize(user_model:, s3_gateway:)
+  def initialize(user_model:, whitelist_checker:, logger: Logger.new(STDOUT))
     @user_model = user_model
-    @s3_gateway = s3_gateway
+    @whitelist_checker = whitelist_checker
+    @logger = logger
   end
 
   def execute(contact:)
     email_address = Mail::Address.new(contact).address
 
-    return unless authorised_email_domain?(email_address)
-
-    login_details = user_model.generate(contact: email_address)
-    send_signup_instructions(email_address, login_details)
+    if whitelist_checker.execute(email_address)[:success]
+      send_signup_instructions(email_address)
+    else
+      logger.info("Unsuccessful email signup attempt: #{email_address}")
+    end
   end
 
 private
 
-  attr_accessor :user_model, :s3_gateway
+  attr_accessor :user_model, :whitelist_checker, :logger
 
-  def authorised_email_domain?(email)
-    regex = Regexp.new(s3_gateway.fetch, Regexp::IGNORECASE)
-    regex.match?(email)
-  end
-
-  def send_signup_instructions(email_address, login_details)
+  def send_signup_instructions(email_address)
     client = Notifications::Client.new(ENV.fetch('NOTIFY_API_KEY'))
 
     client.send_email(
       email_address: email_address,
       template_id: credentials_template_id,
-      personalisation: login_details,
+      personalisation: user_model.generate(contact: email_address),
       email_reply_to_id: do_not_reply_email_address_id
     )
   end
--- a/lib/wifi_user/use_case/sns_notification_handler.rb
+++ b/lib/wifi_user/use_case/sns_notification_handler.rb
@@ -39,7 +39,13 @@ private
     logger.info("Handling signup request from #{from_address}")
 
     ::WifiUser::UseCase::EmailSignup
-      .new(user_model: WifiUser::Repository::User.new)
+      .new(
+        user_model: WifiUser::Repository::User.new,
+        s3_gateway: Common::Gateway::S3ObjectFetcher.new(
+          bucket: ENV.fetch('S3_SIGNUP_WHITELIST_BUCKET'),
+          key: ENV.fetch('S3_SIGNUP_WHITELIST_OBJECT_KEY')
+        )
+      )
       .execute(contact: from_address)
   end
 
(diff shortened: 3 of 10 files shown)


spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
```diff
@@ -8,10 +8,18 @@
...
+  subject do
+    described_class.new(
+      user_model: user_model,
+      whitelist_checker: whitelist_checker
+    )
+  end
 
   before do
     ENV['RACK_ENV'] = environment
+    ENV['NOTIFY_API_KEY'] = 'dummy_key-00000000-0000-0000-0000-000000000000-00000000-0000-0000-0000-000000000000'
```
emileswarts: Noticed this test was failing when run on its own.
Depended on setup from a previous test to set this env variable.

Needs looking into, will do in a separate pr

spec/acceptance/email_notification_spec.rb
```diff
@@ -1,6 +1,4 @@
 RSpec.describe App do
-  before { ENV['AUTHORISED_EMAIL_DOMAINS_REGEX'] = '.gov.uk$' }
```
emileswarts: This did nothing

lib/wifi_user/use_case/sns_notification_handler.rb
```diff
@@ -38,9 +38,15 @@ def handle_signup_request(ses_notification)
...
     logger.info("Handling signup request from #{from_address}")
 
-    ::WifiUser::UseCase::EmailSignup
-      .new(user_model: WifiUser::Repository::User.new)
-      .execute(contact: from_address)
+    ::WifiUser::UseCase::EmailSignup.new(
+      user_model: WifiUser::Repository::User.new,
+      whitelist_checker: WifiUser::UseCases::CheckIfWhitelistedEmail.new(
+        gateway: Common::Gateway::S3ObjectFetcher.new(
+          bucket: ENV.fetch('S3_SIGNUP_WHITELIST_BUCKET'),
```
cybojenix: I'm not too keen on the idea of a use case deciding how a configuration should be loaded.
Ideally the whitelist_checker usecase is initialised outside of the sns-notification-handler

cybojenix: actually thinking about it, I think the email signup use case should also be initialised outside this use case.
While usecases can use one-another, I don't think they should be aware of one another.

spec/unit/lib/wifi_user/use_cases/check_if_whitelisted_email_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      result = subject.execute('')
+      expect(result).to eq(success: false)
+    end
+
+    it 'rejects a nil address' do
+      result = subject.execute(nil)
+      expect(result).to eq(success: false)
+    end
+
+    it 'respects subdomains' do
```
cybojenix: I don't think this test adds anything. The regex above is hard coded to require a `.`, so is just another case of the described class being able to do regex

cybojenix: I do however think a context should be added for testing the regex pattern we're using for subdomains, to make sure the class will work with the upcoming subdomain regex

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
cybojenix: Agreed.
we should do something similar for Env with as we did with the s3 config.
This can act as a reference so we can enforce each example is clean: https://www.rubydoc.info/github/rspec/rspec-support/RSpec%2FSupport%2FShellOut:with_env

lib/wifi_user/use_case/sns_notification_handler.rb
emileswarts: True, app.rb used to be the file that composed and envoked all these classes.

This sns_notification_handler is not a real use case, more of a facade class and I wonder if we need it at all.

Totally agree, once again refactor needed, would suggest separate PR

spec/unit/lib/wifi_user/use_cases/check_if_whitelisted_email_spec.rb
emileswarts: True

lib/wifi_user/use_case/email_signup.rb
```diff
@@ -2,30 +2,33 @@
...
     email_address = Mail::Address.new(contact).address
 
-    return unless Common::EmailAddress.authorised_email_domain?(email_address)
-
-    login_details = user_model.generate(contact: email_address)
-    send_signup_instructions(email_address, login_details)
+    if whitelist_checker.execute(email_address)[:success]
+      send_signup_instructions(email_address)
+    else
+      logger.info("Unsuccessful email signup attempt: #{email_address}")
```
kjdchapman: Is it weird that when a user doesn't have a whitelisted e-mail, we don't reply to them with any help?

That's a feature request though, not for this PR - just thinking out loud.

emileswarts: Yeah getting numbers on this feels like a logical first step

lib/wifi_user/use_case/check_if_whitelisted_email.rb
```diff
@@ -0,0 +1,21 @@
+module WifiUser
+  module UseCases
+    class CheckIfWhitelistedEmail
+      def initialize(gateway:)
+        @gateway = gateway
```
cybojenix: can this be a bit more descriptive as to what is expected, please?
maybe `email_regex_gateway`?

lib/wifi_user/use_case/sns_notification_handler.rb
kjdchapman: Agree with @cybojenix on how composition "should" work - it's ideal for another class to have responsibility for composing things correctly.

I'm still getting my thoughts together on how configuration - especially through environment variables - interacts with the CA constructs. I've used them raw elsewhere, putting behind a gateway feels excessive, passing in from `app.rb` might be correct.

Agree with @emileswarts that any of this work could be done in a follow-up PR.

spec/unit/lib/wifi_user/use_cases/check_if_whitelisted_email_spec.rb
```diff
@@ -0,0 +1,35 @@
+describe WifiUser::UseCases::CheckIfWhitelistedEmail do
+  subject { described_class.new(gateway: s3_gateway) }
+  let(:s3_gateway) { double(fetch: '^.*@(aaa\.uk)$') }
```
cybojenix: `s3` in name, but doesn't relate to `s3` anymore

lib/wifi_user/use_case/sns_notification_handler.rb
kjdchapman: Further thoughts: `handler` as a name could be a smell. The responsibility of this class takes a few reads to understand, and feels a little muddled.

This PR isn't changing that, though - I think the change that _is_ being made is a good one.

spec/acceptance/email_notification_spec.rb
kjdchapman: Is that because this REGEX env-var is set in a Dockerfile somewhere? I don't like specs relying on Dockerfiles, makes the setup distant from the assertion.

spec/unit/lib/wifi_user/use_cases/check_if_whitelisted_email_spec.rb
```diff
@@ -0,0 +1,35 @@
+describe WifiUser::UseCases::CheckIfWhitelistedEmail do
+  subject { described_class.new(gateway: s3_gateway) }
```
kjdchapman: I'd call this `spy_gateway` or similar in a spec.

cybojenix:
Understand that the refactoring has been marked as TODO, so won't block it in this PR

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
kjdchapman: Haven't seen that, looks cool - can you link to where else we took that approach @cybojenix ?

spec/unit/lib/wifi_user/use_cases/check_if_whitelisted_email_spec.rb
emileswarts: Not really a spy though, will call it regex_gateway_stub

emileswarts: If you are cool w that

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
```diff
@@ -148,6 +156,7 @@ def a_plural_sponsor_confirmation_request
     let(:sponsor) { 'adrian <adrian@fake.uk>' }
     let(:sponsees) { ['adrian@notgov.uk'] }
     let(:do_not_reply_id) { staging_do_not_reply_id }
+    let(:whitelist_checker) { double(execute: { success: false }) }
```
kjdchapman: The test structure feels a bit weird to me with the changes to whitelist_checker.

The test reads half like it's an acceptance spec:
- Here's an email that wouldn't work in the real world
- Try out the process
- See behaviour

And half like it's a unit spec with mocking collaborators
- Here's some irrelevant input
- Here's a collaborator that always returns false
- Try out the process
- See behaviour

Maybe some of this setup doesn't need to be here? Or it can be expressed more tersely?

Finding this hard to explain in text, though.

emileswarts: Agreed, once again I don't think this was introduced by this new whitelist feature.

Code needs to be looked at again as a whole, and some classes teased apart.

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
cybojenix: We've done it with a closed source solution in the past, but it would amount to something along the lines of:

```rb
  let(:env_override) do
    { NOTIFY_API_KEY: '' }
  end
  around do |example|
    with_env(env_override) { example.call }
  end
```

PR #147, opened 2019-02-27
Don't hardcode eu-west-1

emileswarts:
The admin bucket is in London.
Also construct the s3 bucket fetcher correctly for the whitelist checker
use case


kjdchapman:
Without it hard-coded, what happens?

emileswarts:
@kjdchapman It blows up, region is required

kjdchapman:
I was confused because this PR is **Don't hardcode eu-west-1** and I'm pretty sure it... hardcodes eu-west-1? :sweat_smile:

PR #150, opened 2019-02-28
Refactor SNSNotificationHandler

emileswarts:
This class is a UseCase with dependencies hardcoded in it, making it
difficult to change.  Inject the SMS and Email handlers and wrap a high
level feature spec around this.

Move NOTIFY_API_KEY to the docker-compose setup as we don't do any
direct assertions against this.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -29,8 +29,28 @@ class App < Sinatra::Base
   end
 
   post '/user-signup/email-notification' do
-    logger.info(request)
-    WifiUser::UseCase::SnsNotificationHandler.new(logger).handle(request)
+    whitelist_checker = WifiUser::UseCases::CheckIfWhitelistedEmail.new(
+      gateway: Common::Gateway::S3ObjectFetcher.new(
+        bucket: ENV.fetch('S3_SIGNUP_WHITELIST_BUCKET'),
+        key: ENV.fetch('S3_SIGNUP_WHITELIST_OBJECT_KEY'),
+        region: 'eu-west-2'
+      )
+    )
+
+    email_signup_handler = ::WifiUser::UseCase::EmailSignup.new(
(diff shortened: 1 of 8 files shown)


app.rb
```diff
@@ -29,8 +29,28 @@ class App < Sinatra::Base
   end
 
   post '/user-signup/email-notification' do
-    logger.info(request)
-    WifiUser::UseCase::SnsNotificationHandler.new(logger).handle(request)
+    whitelist_checker = WifiUser::UseCases::CheckIfWhitelistedEmail.new(
```
cybojenix: I think these can be lifted even higher, and moved to the app initialisation phase instead of being part of the controller.

The only thing tied to the controller is the final execution.
Feel it will make controller testing easier (app configuration can be modified instead of patching the classes), and also improve efficiency (only initialised once)

app.rb
```diff
@@ -29,8 +29,28 @@ class App < Sinatra::Base
...
+      gateway: Common::Gateway::S3ObjectFetcher.new(
+        bucket: ENV.fetch('S3_SIGNUP_WHITELIST_BUCKET'),
```
cybojenix: Would avoid letting a controller use `ENV`, and would lean towards a controller using the sinatra's settings instead.

However with the above comment, I don't think this will apply anymore

app.rb
emileswarts: Given this is a refactoring, I'm not making any improvements just yet.
Happy to have a discussion, also have some ideas around this.

app.rb
emileswarts: Sounds good but propose this be done independent of this refactoring step

app.rb
cybojenix: sure thing

PR #151, opened 2019-02-28
Message parser

emileswarts:
This is currently being done by the sns notification handler (which has
many responsibilities already).

Isolate this ugly JSON request message parsing.

This class is unused right now but will be injected as a dependency into
the SNSNotificationHandler in a future PR.

### DIFF
--- a/lib/wifi_user/use_case/parse_email_request.rb
+++ b/lib/wifi_user/use_case/parse_email_request.rb
@@ -1,0 +1,24 @@
+class WifiUser::UseCase::ParseEmailRequest
+  def initialize(logger: Logger.new(STDOUT))
+    @logger = logger
+  end
+
+  def execute(request)
+    parsed_request = JSON.parse(request)
+    parsed_message = JSON.parse(parsed_request.fetch('Message'))
+    logger.debug("Processing request: #{parsed_request} with message #{parsed_message}")
+
+    {
+      type: parsed_request['Type'],
+      message_id: parsed_message['mail']['messageId'],
+      from_address: parsed_message['mail']['commonHeaders']['from'][0],
+      to_address: parsed_message['mail']['commonHeaders']['to'][0],
+      s3_object_key: parsed_message['receipt']['action']['objectKey'],
+      s3_bucket_name: parsed_message['receipt']['action']['bucketName']
+    }
+  end
+
+private
+
+  attr_reader :logger
+end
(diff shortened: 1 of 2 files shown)


lib/wifi_user/use_case/parse_email_request.rb
```diff
@@ -0,0 +1,24 @@
+class WifiUser::UseCase::ParseEmailRequest
+  def initialize(logger: Logger.new(STDOUT))
+    @logger = logger
+  end
+
+  def execute(request)
+    parsed_request = JSON.parse(request)
+    parsed_message = JSON.parse(parsed_request.fetch('Message'))
+    logger.info("Processing request: #{parsed_request} with message #{parsed_message}")
```
cybojenix: might consider this to be `debug` instead of `info`

spec/unit/lib/wifi_user/use_cases/parse_email_request_spec.rb
```diff
@@ -0,0 +1,21 @@
+describe WifiUser::UseCase::ParseEmailRequest do
+  describe 'parsing the request' do
+    it 'returns the type when one is present' do
+      request = '{
+        "Type":"Notification",
+        "Message":"{\"mail\":{\"messageId\":\"123\",\"commonHeaders\":{\"from\":[\"bob@gov.uk\"],\"to\":[\"sally@gov.uk\"]}},\"receipt\":{\"action\":{\"objectKey\":\"some-object-key\",\"bucketName\":\"some-bucket-name\"}}}"
+      }'
```
emileswarts: Realise this is pretty awful to read but this is the input to our system.
Not sure if converting this to and from something more readable is an unneeded layer of indirection.

cybojenix: agree, let's stick with this. If it turns out modifying this is a pain point, we can change to encoding a plain object.

kjdchapman: Oof, yeah, probably doesn't actually help much to introduce the indirection, especially as we don't repeat this pattern (afaik).

This is a fairly small spec file - maybe it looks better spaced out a bit?

e.g.
```
"Message":"{
  \"mail\":{
    \"messageId\":\"123\",
    \"commonHeaders\":{
      \"from\":[\"bob@gov.uk\"],
      \"to\":[\"sally@gov.uk\"]}},
      \"receipt\":{
        \"action\":{
          \"objectKey\":\"some-object-key\",\"bucketName\":\"some-bucket-name\"
        }
      }
    }"
```

... still kind of horrible

cybojenix: unfortunately you can't do that, as you cu'an't have multiple lines in JSON.
You'd have to escape the line endings too

PR #152, opened 2019-02-28
Use new email parser in SNS notification handler

emileswarts:
The knowlege of the request is now being dealt with by the email parser.
We now process the entire request before handling it, which forces some
tests to change.
Use fetch instead of square brackets so we can raise a key error instead
of a nomethod error ([] on nil).  This can be rescued and we can short
curcuit execution.  At this point we will log to Cloudwatch.

### DIFF
--- a/lib/wifi_user/use_case/parse_email_request.rb
+++ b/lib/wifi_user/use_case/parse_email_request.rb
@@ -9,16 +9,36 @@ class WifiUser::UseCase::ParseEmailRequest
     logger.debug("Processing request: #{parsed_request} with message #{parsed_message}")
 
     {
-      type: parsed_request['Type'],
-      message_id: parsed_message['mail']['messageId'],
-      from_address: parsed_message['mail']['commonHeaders']['from'][0],
-      to_address: parsed_message['mail']['commonHeaders']['to'][0],
-      s3_object_key: parsed_message['receipt']['action']['objectKey'],
-      s3_bucket_name: parsed_message['receipt']['action']['bucketName']
+      type: parsed_request.fetch('Type'),
+      message_id: message_id(parsed_message),
+      from_address: from_address(parsed_message),
+      to_address: to_address(parsed_message),
+      s3_object_key: s3_object_key(parsed_message),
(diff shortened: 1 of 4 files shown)


spec/acceptance/email_notification_spec.rb
```diff
@@ -52,6 +54,14 @@ def post_notification
         post_notification
         expect(last_response.body).to eq('')
       end
+
+      describe 'POSTing a Amazon SES Setup Notification to /user-signup/email-notification' do
```
kjdchapman: I think the wrapping block already gives most of the context, so this could be terser?

```suggestion
      describe 'of the "Amazon SES Setup" type' do
```

lib/wifi_user/use_case/parse_email_request.rb
```diff
@@ -9,16 +9,36 @@ def execute(request)
...
+      s3_object_key: s3_object_key(parsed_message),
+      s3_bucket_name: s3_bucket_name(parsed_message)
     }
   end
 
 private
 
   attr_reader :logger
+
+  def message_id(request)
```
kjdchapman: Interestingly we just took the opposite approach in another PR - in-lining a bunch of one-line methods.

I think this is the right choice here, though. Hide some of this nested fetch horribleness.

lib/wifi_user/use_case/parse_email_request.rb
```diff
@@ -9,16 +9,36 @@ def execute(request)
...
+    request.fetch('mail').fetch('messageId')
+  end
+
+  def from_address(request)
+    request.fetch('mail').fetch('commonHeaders').fetch('from').fetch(0)
```
cybojenix: raising that the rationale for `fetch` was to catch the `KeyError`, however `Array#fetch` will raise an `IndexError`

spec/acceptance/email_notification_spec.rb
cybojenix: Aware this is from legacy, but would rather a whitelist be used as apposed to a black list.
IE, check for the messageId we want, not the one we don't care about

PR #154, opened 2019-03-01
Not allow vowels within usernames

georgeschena:


spec/unit/lib/wifi_user/repository/user_spec.rb
```diff
@@ -46,6 +46,18 @@
     end
   end
 
+  context 'avoiding usernames including aeiou' do
```
emileswarts: Might be good to say why we are doing this

context 'avoiding potentially offensive random words' do
    it 'generates usernames without vowels' do

PR #157, opened 2019-03-04
implement basic validation to ensure our SNS handler only deals with notifications

cybojenix:
This is required as we don't use handle subscription confirmations yet

### DIFF
--- a/lib/wifi_user/use_case/sns_notification_handler.rb
+++ b/lib/wifi_user/use_case/sns_notification_handler.rb
@@ -7,6 +7,10 @@ class WifiUser::UseCase::SnsNotificationHandler
   end
 
   def handle(request)
+    if request_invalid?(request)
+      logger.warn("Unexpected request: \n #{request.body.read}")
+      return
+    end
     params = request.body.read
 
     begin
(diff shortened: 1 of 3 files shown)


lib/wifi_user/use_case/sns_notification_handler.rb
```diff
@@ -56,4 +60,17 @@ def handle_sponsor_request(payload)
...
+    !request_valid?(request)
+  end
+
+  def request_valid?(request)
+    # For now, we only care that the correct header is set to see if we're
+    # actually dealing with a notification.
+    # There is much more that should be in here.
+
+    request.has_header?('x-amz-sns-message-type') \
+    && request.get_header('x-amz-sns-message-type') == 'Notification'
```
emileswarts: Was aiming to push the knowledge of the internals of this request down into the parser object I created.
Feels like we need a domain object that has a known interface that we can return to this use case?

Happy to discuss.

emileswarts:
I am happy for this to go in now, would like to spend another refactoring session on this though.

cybojenix:
noting we had an offline discussion for how this will be implemented correctly, which is to be done before we set up new domains.

For now, this lets us get staging working again in a safe manner.

PR #158, opened 2019-03-04
Allow letters to be used more than once in a username

georgeschena:


lib/wifi_user/repository/user.rb
```diff
@@ -26,7 +26,7 @@ def random_username
   end
 
   def generate_username
-    ALPHABET_WITHOUT_VOWELS.sample(6).join
+    (0...6).map { ALPHABET_WITHOUT_VOWELS[rand(6)] }.join
```
emileswarts: Will this ever touch letters beyond "h"?
This 6 looks like it's going to only use a small subset of our alphabet.

georgeschena: Oh yeh my bad!!

lib/wifi_user/repository/user.rb
```diff
@@ -26,7 +26,7 @@ def random_username
   end
 
   def generate_username
-    ALPHABET_WITHOUT_VOWELS.sample(6).join
+    (0...6).map { ALPHABET_WITHOUT_VOWELS[rand(21)] }.join
```
emileswarts: Would be nice to have the upper bound dynamic instead of a hardcoded 21.

Could do:
    (0...6).map { ALPHABET_WITHOUT_VOWELS[rand(ALPHABET_WITHOUT_VOWELS.count)] }.join

Or even this would achieve the same effect:
ALPHABET_WITHOUT_VOWELS.shuffle[0...6].join

PR #159, opened 2019-03-07
Use a gateway for sending SMS via gov notify

cybojenix:


lib/wifi_user/gateway/gov_notify_sms.rb
```diff
@@ -0,0 +1,20 @@
+require 'notifications/client'
+
+class WifiUser::Gateway::GovNotifySMS
+  def initialize(api_key, template_id)
+    @client = Notifications::Client.new(api_key)
+    @template_id = template_id
```
cybojenix: I'm torn between this being at initialisation or execution.

On the one hand, it allows a more static class structure if each SMS template we have as a counter part gateway:

```
send_credentials_sms = GovNotifySMS(api_key, config['notify_sms_template_ids']['credentials'])
```
but then this will result in needing to restructure the `sms_template_finder`.

Think by writing this, I've realised that there should be one gateway and multiple use-cases.

lib/wifi_user/domain/sms_response.rb
```diff
@@ -0,0 +1,7 @@
+class WifiUser::Domain::SMSResponse
```
kjdchapman: I wouldn't introduce a domain object until it hits some level of complexity - I'd just pass around a hash, even from a gateway. 

I might introduce it where I know complexity is imminent, too - is this just the first slice of an `sms_response`?

cybojenix: yes. The next piece of this is to add error handling.
I'd rather use a domain to define a static structure, even for something as simple as this.
Use of hashes very quickly leads to implicit complexity if left unchecked

Makefile
```diff
@@ -31,6 +31,10 @@ test:
 	$(DOCKER_COMPOSE) run --rm app rspec
 	$(MAKE) stop
 
+shell: serve
+	./mysql/bin/wait_for_mysql
+	$(DOCKER_COMPOSE) run --rm app ash
```
emileswarts: ash?

cybojenix: it's running alpine, so no bash

lib/wifi_user/domain/sms_response.rb
kjdchapman: Cool, happy with this then.

spec/unit/lib/wifi_user/gateway/gov_notify_sms_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    it { expect(subject.success).to be true }
+  end
+
+  # These contexts are to test the different specified errors from the API
+  # https://docs.notifications.service.gov.uk/ruby.html#error-codes
+
+  context 'with unexpected error' do
+    # This will cause a parse error, which we want to see as an exception
+    let(:return_body) { '' }
+    it { expect { subject }.to raise_error(Exception) }
```
kjdchapman: Can we pick out a more specific exception here?

lib/wifi_user/gateway/gov_notify_sms.rb
```diff
@@ -0,0 +1,31 @@
+require 'notifications/client'
+
+class WifiUser::Gateway::GovNotifySMS
+  def initialize(api_key)
+    @client = Notifications::Client.new(api_key)
+  end
+
+  def execute(phone_number, template_id, template_parameters: {})
```
kjdchapman: Decided to stick with template_id in the execute? 

Agreed that it felt weird, but don't feel strongly about moving it right this minute.

cybojenix: yeah, it'll cause a lot of mess as currently most template IDs are fetched from the template finder.

It might be that at some point, the template finder becomes something that dishes out gateways, but that's for the future when this project gets refactored

spec/unit/lib/wifi_user/gateway/gov_notify_sms_spec.rb
cybojenix: I guess the reason this tests exists is to demonstrate that we still want to propagate exceptions, and only deal with BadRequestError only.

Do you have any suggests for how better to demonstrate?

kjdchapman: I would explicitly test the type of exception I'm expecting to bubble out - I forget what this is called in Ruby, is that the `BadRequestError` in this case? - because I'd be worried about false positives on this test.

If I wanted to demonstrate it doesn't capture arbitrary errors, I'd probably just provide three or four examples - possibly throwing errors from a mocked collaborator. I don't feel like that's necessary here, though.

cybojenix: ok, will remove.

This was done before I narrowed the scope of what errors would be caught, so definitely less important

PR #160, opened 2019-03-07
Wire up Sponsors use case to use the new send sms gateway

cybojenix:


app.rb
```diff
@@ -14,6 +14,10 @@ class App < Sinatra::Base
 
   configure do
     set :log_level, Logger::DEBUG
+    
+    set :gateways, {
+      send_sms: WifiUser::Gateway::GovNotifySMS.new(ENV.fetch('NOTIFY_API_KEY'))
+    }
```
emileswarts: Looks like this is moving quite quickly towards a usecase factory.
While I'm not against this, we've not done this on any other application.

Wonder if we should defer this until it becomes a pain to manage the dependencies.

app.rb
```diff
@@ -14,6 +14,9 @@ class App < Sinatra::Base
 
   configure do
     set :log_level, Logger::DEBUG
+
+    set :gateways,
+      send_sms: WifiUser::Gateway::GovNotifySMS.new(ENV.fetch('NOTIFY_API_KEY'))
```
cybojenix: for my peace of mind (+ others unsure), the use cases can depend on the gateways by doing this:

```ruby
set :use_cases, -> do
  {
    some_usecase: SomeUsecase.new(gw: settings.gateways.some_gateway)
  }
end
```

app.rb
cybojenix: I get the concern.
This is to avoid adding more responsibility to the controller. Given that it's still within the file, and is more or less the same as how we've been doing it elsewhere (it's just assigned to a context variable instead of a scoped variable), then this should be fine

PR #161, opened 2019-03-08
report SMS sponsee failures to the sponsor

cybojenix:
**WHAT**
report to the sponsor any of the contact details that failed being sent (for phone numbers)

**WHY**
Currently if GovNotify reports a validation error to us, it raises an exception, and the sponsor isn't told anything.
Give the sponsor the opportunity to correct any invalid data in the event of a failure.

**NOTE**
this does not cover reporting validation on email addresses, or dropping sponsees within our contact sanitiser

### DIFF
--- a/lib/wifi_user/use_case/sponsor_users.rb
+++ b/lib/wifi_user/use_case/sponsor_users.rb
@@ -72,7 +72,7 @@ private
   end
 
   def send_confirmation_email(sponsor, sponsees, failed_sponsees: [])
-    return send_sponsor_failed(sponsor, failed_sponsees: failed_sponsees) if sponsees.empty? || !failed_sponsees.empty?
+    return send_failed_sponsoring_email(sponsor, failed_sponsees: failed_sponsees) if sponsees.empty? || !failed_sponsees.empty?
     return send_confirmation_singular(sponsor, sponsees) if sponsees.length == 1
 
     send_confirmation_plural(sponsor, sponsees)
@@ -101,7 +101,7 @@ private
     )
   end
 
-  def send_sponsor_failed(sponsor_address, failed_sponsees: [])
+  def send_failed_sponsoring_email(sponsor_address, failed_sponsees: [])
     notify_client.send_email(
       email_address: sponsor_address,
       template_id: sponsor_confirmation_template['failed'],
--- a/lib/wifi_user/use_case/sponsor_users.rb
+++ b/lib/wifi_user/use_case/sponsor_users.rb
@@ -12,8 +12,8 @@ class WifiUser::UseCase::SponsorUsers
 
     if whitelist_checker.execute(sponsor_address)[:success]
       sponsees = sanitise_sponsees(unsanitised_sponsees)
-      invite_sponsees(sponsor, sponsor_address, sponsees)
-      send_confirmation_email(sponsor_address, sponsees)
+      failed_sponsees = invite_sponsees(sponsor, sponsor_address, sponsees)
+      send_confirmation_email(sponsor_address, sponsees, failed_sponsees: failed_sponsees)
     else
       logger.info("Unsuccessful sponsor signup attempt: #{sponsor_address}")
     end
@@ -28,7 +28,7 @@ private
   end
 
   def invite_sponsees(sponsor, sponsor_address, sponsees)
-    sponsees.each do |sponsee|
+    failed_sponsees = sponsees.reject do |sponsee|
       if email?(sponsee)
         sponsee_address = Mail::Address.new(sponsee).address
         sponsor_email(sponsor, sponsor_address, sponsee_address)
@@ -36,6 +36,7 @@ private
         sponsor_phone_number(sponsor_address, sponsee)
       end
     end
+    failed_sponsees
   end
 
   def notify_client
@@ -51,7 +52,7 @@ private
         login: login_details[:username],
         pass: login_details[:password]
       }
-    )
+    ).success
   end
 
   def sponsor_email(sponsor, sponsor_address, sponsee_address)
@@ -62,14 +63,16 @@ private
       personalisation: login_details.merge(sponsor: sponsor),
       email_reply_to_id: do_not_reply_email_address_id
     )
+    # FIXME: we currently don't support checking if the email was invalid, so always say it is
+    true
   end
 
   def config
     YAML.load_file("config/#{ENV['RACK_ENV']}.yml")
   end
 
-  def send_confirmation_email(sponsor, sponsees)
-    return send_sponsor_failed(sponsor) if sponsees.empty?
+  def send_confirmation_email(sponsor, sponsees, failed_sponsees: [])
+    return send_sponsor_failed(sponsor, failed_sponsees: failed_sponsees) if sponsees.empty? || !failed_sponsees.empty?
     return send_confirmation_singular(sponsor, sponsees) if sponsees.length == 1
 
     send_confirmation_plural(sponsor, sponsees)
@@ -98,11 +101,13 @@ private
     )
   end
 
-  def send_sponsor_failed(sponsor_address)
+  def send_sponsor_failed(sponsor_address, failed_sponsees: [])
     notify_client.send_email(
       email_address: sponsor_address,
       template_id: sponsor_confirmation_template['failed'],
-      personalisation: {},
+      personalisation: {
+        failedSponsees: format_failed_sponsees(failed_sponsees)
+      },
       email_reply_to_id: do_not_reply_email_address_id
     )
   end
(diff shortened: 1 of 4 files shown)


lib/wifi_user/use_case/sponsor_users.rb
```diff
@@ -12,8 +12,8 @@ def execute(unsanitised_sponsees, sponsor)
 
     if whitelist_checker.execute(sponsor_address)[:success]
       sponsees = sanitise_sponsees(unsanitised_sponsees)
-      invite_sponsees(sponsor, sponsor_address, sponsees)
-      send_confirmation_email(sponsor_address, sponsees)
+      failed_sponsees = invite_sponsees(sponsor, sponsor_address, sponsees)
```
kjdchapman: I'm surprised by an `invite_sponsees` method returning a naked `failed_sponsees` - but the naming does make it clear what happens from here on out. Hmm.

lib/wifi_user/use_case/sponsor_users.rb
```diff
@@ -28,14 +28,15 @@ def sanitise_sponsees(contacts)
   end
 
   def invite_sponsees(sponsor, sponsor_address, sponsees)
-    sponsees.each do |sponsee|
+    failed_sponsees = sponsees.reject do |sponsee|
```
kjdchapman: I'd be happy without assigning a variable here, feels more "ruby". I think it's clear what we're doing without the name? I'm 60-40 though.

`invite_and_return_failed_sponsees` isn't good naming... right?

lib/wifi_user/use_case/sponsor_users.rb
```diff
@@ -62,14 +63,16 @@ def sponsor_email(sponsor, sponsor_address, sponsee_address)
       personalisation: login_details.merge(sponsor: sponsor),
       email_reply_to_id: do_not_reply_email_address_id
     )
+    # FIXME: we currently don't support checking if the email was invalid, so always say it is
```
kjdchapman: Is this for a later PR?

kjdchapman: Not sure how I feel about FIXME's ending up on `master`. Feel like we've not had that discussion as a team ever!

lib/wifi_user/use_case/sponsor_users.rb
```diff
@@ -98,11 +101,13 @@ def send_confirmation_singular(sponsor_address, sponsees)
     )
   end
 
-  def send_sponsor_failed(sponsor_address)
+  def send_sponsor_failed(sponsor_address, failed_sponsees: [])
```
kjdchapman: I would call this `send_failed_sponsoring_email`, or something like that. 

Think there's the opportunity to make this more clear through naming, at least.

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
```diff
@@ -146,8 +146,11 @@ def a_plural_sponsor_confirmation_request
     it 'Sends a sponsorship failed email to the sponsor' do
```
kjdchapman: I know this is pre-existing, but nit: casing

```suggestion
    it 'sends a sponsorship failed email to the sponsor' do
```

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
```diff
@@ -165,6 +168,76 @@ def a_plural_sponsor_confirmation_request
     end
   end
 
+  context 'On Failing to send an SMS' do
```
kjdchapman: nit: casing
```suggestion
  context 'on failing to send an SMS' do
```

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
```diff
@@ -146,8 +146,11 @@ def a_plural_sponsor_confirmation_request
     it 'Sends a sponsorship failed email to the sponsor' do
       body = {
         email_address: 'cassandra@gov.uk',
-        template_id: 'efc83658-dcb5-4401-af42-e26b1945c1a9',
-        personalisation: {},
+        template_id: '52c19b68-4d8b-497a-b6ae-ee27d49439c3',
+        personalisation: {
+          # TODO: this should pick up failed sponsees in the contact sanitisor
```
kjdchapman: Same comment as the FIXME above.

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
```diff
@@ -165,6 +168,76 @@ def a_plural_sponsor_confirmation_request
...
 
+  context 'On Failing to send an SMS' do
+    let(:sponsor) { 'Cassandra <cassandra@gov.uk>' }
+    let(:success_sponsees) { [] }
+    let(:failed_sponsees) { %w(+447770000666) }
+    let(:sponsees) { success_sponsees + failed_sponsees }
+    let(:do_not_reply_id) { production_do_not_reply_id }
+    let(:formatted_failed_sponsees) { "* +447770000666" }
+
+    let(:send_sms_gateway) do
```
kjdchapman: You're right that this is a monster of a mock. 

A few thoughts:
- In general, I think it's the right decision to mock here - we have control over this collaborator and it's a unit spec.
- Is this mock doing too much? It's both _verifying what it receives_ and _giving specific responses_ - does that make reading it difficult?
- Is hiding this mock behind a level of abstraction valuable? It'll make the spec more readable at a high-level, but might make it harder to understand this specific part.

lib/wifi_user/use_case/sponsor_users.rb
cybojenix: yeah, maybe this should return an object instead.
```
def invite_sponsees(sponsor, sponsor_address, sponsees) -> Hash[:succeeded, :failed]
```

lib/wifi_user/use_case/sponsor_users.rb
cybojenix: see above, maybe going the hash route is better?

I didn't like `sponsees.reject` on it's own, as it tells us nothing about what's happening

lib/wifi_user/use_case/sponsor_users.rb
cybojenix: there definitely needs to be some comment about why this returns true.

Starting to see that we should update the signature of all theses affected functions to make it clearer that they all now cascade upwards to report failures.

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
cybojenix:
```suggestion
    it 'sends a sponsorship failed email to the sponsor' do
```

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
cybojenix: yeah agree with this one outright, as it won't cause confusion if seen

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
cybojenix: would like to make clear that this double isn't being used as a mock. We aren't verifying any executions, and only apply special behaviour with certain fixed inputs (essentially it's a Fake, not a Mock).
It allows behaviour, however only to the same degree the mocks are above.

I'm going to lean on saying yes, the abstraction is very valuable.
It's contained to a context, and now each of the sub-contexts are responsible only for what they care about, which is to verify combinations of successful and unsuccessful calls are made.

I'm open to suggestions for how to do this better with the abstraction in place.
Perhaps the setup can be broken out into a function, and placed in the `before` block?

```rb
let(:send_sms_gateway) { double(execute: double(success: true)) }

before(:each) do
  set_send_sms_gateway_execution_branches(
    successful_numbers: success_sponsees, unsucessful_numbers: failed_sponsees
  )
end

def set_send_sms_gateway_execution_branches(successful_numbers:, unsucessful_numbers:)
  ...
end
```

lib/wifi_user/use_case/sponsor_users.rb
georgeschena: Yeah I don't like the FIXME's Im down to get rid

lib/wifi_user/use_case/sponsor_users.rb
georgeschena: Yeah I would rather not use `sponsees.reject` doesn't describe the issue well imo.

georgeschena: Better to assign it to a nice variable

spec/unit/lib/wifi_user/use_cases/sponsor_users_spec.rb
georgeschena: > Perhaps the setup can be broken out into a function, and placed in the before block? 
I like this

PR #167, opened 2019-03-12
Fix sentry bad request error

Renny4Real:
**WHY:**
Sentry Error: 
`Notifications::Client::BadRequestError: [{"error"=>"ValidationError", "message"=>"phone_number None is not of type string"}]`

We have discovered this is due to a named number (like what you get from a bank, where it says LLOYDS as the sender, instead of a number). These cannot be replied to.

**IN THIS PR:**
- Recreate the cause of the error with tests.
- Stop the use case from generating a message based on these kinds of credentials.
- Add a logging functionality that logs to Cloudwatch when this has occurred.

**Any suggestions/feedback would be appreciated**

### DIFF
--- a/spec/unit/lib/wifi_user/use_cases/sms_response_spec.rb
+++ b/spec/unit/lib/wifi_user/use_cases/sms_response_spec.rb
@@ -3,7 +3,7 @@ describe WifiUser::UseCase::SmsResponse do
   let(:template_finder) { double(execute: notify_template_id) }
   subject { described_class.new(user_model: user_model, template_finder: template_finder) }
 
-  context 'With invalid credentials' do
+  context 'With named number' do
     let(:phone_number) { 'HIDDENNUMBER' }
     let(:notify_template_id) { '00000000-7777-8888-9999-000000000000' }
     let(:notify_sms_url) { stub_request(:post, 'https://api.notifications.service.gov.uk/v2/notifications/sms') }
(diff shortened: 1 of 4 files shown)


spec/unit/lib/wifi_user/use_cases/sms_response_spec.rb
```diff
@@ -3,74 +3,92 @@
   let(:template_finder) { double(execute: notify_template_id) }
   subject { described_class.new(user_model: user_model, template_finder: template_finder) }
 
-  before do
-    expect(user_model).to receive(:generate).with(contact: phone_number).and_return(username: username, password: password)
-  end
+  context 'With invalid credentials' do
```
cybojenix:
```suggestion
  context 'With named number' do
```

spec/unit/lib/wifi_user/use_cases/sms_response_spec.rb
```diff
@@ -3,74 +3,92 @@
...
+    it 'does not send a message to this number' do
+      subject.execute(contact: phone_number, sms_content: '')
+      expect(notify_sms_url).not_to have_been_requested
+    end
   end
 
-  it 'Creates a user prepended by +44' do
-    subject.execute(contact: '07700900003', sms_content: '')
-  end
+  context 'with valid credentials'do
```
cybojenix:
```suggestion
  context 'With valid phone number' do
```

PR #168, opened 2019-03-12
add the gov notify email gateway

cybojenix:


Renny4Real:
Looks good. Is this to fix the sentry error?

cybojenix:
Yes, this is so we can cleanly report issues up to the use case. Same pattern we took with the SMS gateway

PR #170, opened 2019-03-13
add a quick and dirty mechanism for preventing message loops from happening

cybojenix:
We recently had an incident where we were stuck in a loop due a number texting itself.


cybojenix:
Test definitely needs adding.
Will do this after I've confirmed format of the destination number (international vs local)

PR #179, opened 2019-03-26
Use Gov.uk Notify alongside Firetext

tommotaylor:
This PR allows us to receive inbound SMS messages via Notify alongside receiving them from Firetext. It is the first step towards moving to Notify and removing Firetext.

In summary:
- New API endpoint for inbound Notify SMSs
- New endpoint is secured by bearer tokens
- Adds token security to existing Firetext endpoint


app.rb
```diff
@@ -87,6 +87,31 @@ class App < Sinatra::Base
     ''
   end
 
+  post '/user-signup/sms-notification/notify' do
+    source = params[:source_number]
+    destination = params[:destination_number]
+    message = params[:message]
+    logger.info("Processing SMS on /user-signup/sms-notification/notify from #{source} to #{destination} with message #{params[:message]}")
```
tommotaylor: Replace `params[:message]` with `message`

spec/features/notify/sms_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+              it_behaves_like "rejecting an SMS"
+            end
+          end
+        end
+      end
+    end
+
+    def created_user
+      WifiUser::Repository::User.find(contact: internationalised_phone_number)
+    end
```
tommotaylor: this `def created_user` should be a `let` should it not?

cybojenix: this is a copy + paste of the firetext. but yes, will quickly do that now

PR #199, opened 2019-05-29
Append whitespace to username for SMS notifications

steventux:
https://trello.com/c/opNiPkUn/22-no-space-between-end-of-username-and-password-on-end-user-signup-sms

Some mobile devices do not present the login and password credentials on separate lines n the SMS response.
This leads to the username being concatenated with the SMS template text preceding the password. This often confuses users as they attempt to transfer these values into their Wifi settings.

This commit appends a single space char to the end of the login value for clearer presentation in the SMS template. Ideally we'd amend the template to ensure the correct layout but these managed in Notify and would be harder to test.


steventux:
Closing as amending templates is preferable.

PR #202, opened 2019-06-07
Strip spaces from phone numbers

steventux:
https://trello.com/c/i8AXvykM/40-when-sponsoring-a-phone-number-that-contains-a-space-it-doesnt-sponsor-that-number

Remove spaces from sponsee phone numbers before saving and notifying them via SMS.

### DIFF
--- a/lib/wifi_user/use_case/contact_sanitiser.rb
+++ b/lib/wifi_user/use_case/contact_sanitiser.rb
@@ -4,7 +4,7 @@ class WifiUser::UseCase::ContactSanitiser
   def execute(contact)
     first_match(
       email_match(contact) ||
-      internationalize(phone_match(contact)) ||
+      internationalize(phone_match(contact.delete(' '))) ||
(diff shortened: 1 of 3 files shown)


lib/wifi_user/use_case/contact_sanitiser.rb
```diff
@@ -30,6 +30,6 @@ def first_match(match)
   def internationalise_phone_number(phone_number)
     phone_number = '44' + phone_number[1..-1] if phone_number[0..1] == '07'
     phone_number = '+' + phone_number unless phone_number[0] == '+'
-    phone_number
+    phone_number.delete(' ')
```
cybojenix: Given this isn't related to internationalisation, it might be best to break this out to a different function.

Perhaps in a way so the regex doesn't need changing?
```
internationalize(phone_match(strip(contact))) 
```

steventux:
@cybojenix I've simplified this.

PR #203, opened 2019-06-07
Filter sponsor email address when extracting sponsees

steventux:
https://trello.com/c/1NR2y1XC/41-sponsors-emails-eg-in-signature-should-not-themselves-be-sponsored

Filters sponsors email from the list of extracted sponsee addresses.

### DIFF
--- a/lib/wifi_user/use_case/email_sponsees_extractor.rb
+++ b/lib/wifi_user/use_case/email_sponsees_extractor.rb
@@ -1,16 +1,16 @@
 require 'nokogiri'
 
 class WifiUser::UseCase::EmailSponseesExtractor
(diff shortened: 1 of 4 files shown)


lib/wifi_user/use_case/email_sponsees_extractor.rb
```diff
@@ -1,14 +1,16 @@
 require 'nokogiri'
 
 class WifiUser::UseCase::EmailSponseesExtractor
-  def initialize(email_fetcher:)
+  def initialize(email_fetcher:, sponsor_address:)
```
cybojenix: to keep this class contained, might be worth renaming this to be more generalised:

`exluded_sponsees`

PR #209, opened 2019-06-13
Fix wordlist seed during build

phantas:
Fixes an issue where we were not retrieving the wordlist seed file at build time.

TODO: refactor to use aws-helpers in govwifi-concourse-deploy-pipelines

Co-Authored-By: Steve Laing <steve.laing@digital.cabinet-office.gov.uk>

### DIFF
--- a/ci/tasks/build-deployable.yml
+++ b/ci/tasks/build-deployable.yml
@@ -28,5 +28,6 @@ run:
   args:
     - '-exc'
     - |
+      aws s3 cp s3://govwifi-wordlist/wordlist-short ./tmp/wordlist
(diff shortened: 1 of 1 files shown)


ci/tasks/build-deployable.yml
```diff
@@ -28,5 +28,6 @@ run:
   args:
     - '-exc'
     - |
+      aws s3 s3://govwifi-wordlist/wordlist-short ./tmp/wordlist
```
whpearson: This needs to be `aws s3 cp`

phantas: click suggest next time :)

PR #210, opened 2019-06-14
RED-104: Pass wordlist file as an input to container build task

rjbaker:
The build container does not have the aws cli installed. Instead, the deploy pipeline has been updated to download the wordlist file resource which may now be accessed within this task via an input.

### DIFF
--- a/ci/tasks/build-deployable.yml
+++ b/ci/tasks/build-deployable.yml
@@ -16,6 +16,7 @@ params:
 
 inputs:
   - name: src
+  - name: wordlist-file
 
 outputs:
   - name: image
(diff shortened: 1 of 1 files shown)


ci/tasks/build-deployable.yml
```diff
@@ -16,6 +16,7 @@ params:
 
 inputs:
   - name: src
+  - name: wordlist-file
```
phantas: I may be missing something, where is the wordlist-file coming from?

paroxp: https://github.com/alphagov/govwifi-concourse-deploy-pipeline/pull/3/files

rjbaker: This is an input to the task from the s3 resource defined in alphagov/govwifi-concourse-deploy-pipeline#3

phantas: Seen it. I suppose there is no way of testing till both pipelines are in master? I would like to see it running but since the pipeline is broken right now, I am happy to approve and roll back if needed.

PR #268, opened 2019-10-28
Add batch deletion of old user accounts to reduce RDS lag

smford:
Currently when the deletion of old user accounts occur, a lag
in RDS replication occurs, which in turn fires a cloudwatch
alarm. As it self heals, this alarm is ignored but it does
create unnecessary noise which we have all become numb to.

We believe the cause is because of how large the query is,
and that it creates a single large transaction log which
needs to be first copied across and fully replayed on the
replica, and during this replay the lag is created.

This PR breaks the old user account deletion query into
batches of 500, with the intent to allow each batch to be
replicated and replayed on to the replica whilst other
batches are still being created on the master, thus
dampening the effect of the replication and reducing lag.

### DIFF
--- a/lib/gdpr/gateway/usedetails.rb
+++ b/lib/gdpr/gateway/usedetails.rb
@@ -8,7 +8,7 @@ class Gdpr::Gateway::Userdetails
 
     total = 0
     loop do
-      deleted_rows = DB[:sessions].with_sql_delete("
+      deleted_rows = DB[:userdetails].with_sql_delete("
         DELETE FROM userdetails WHERE (last_login < DATE_SUB(NOW(), INTERVAL 12 MONTH)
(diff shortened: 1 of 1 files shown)


lib/gdpr/gateway/usedetails.rb
```diff
@@ -1,9 +1,26 @@
...
+  SESSION_BATCH_SIZE = 500
   def delete_users
-    DB.run('DELETE FROM userdetails
-        WHERE (last_login < DATE_SUB(NOW(), INTERVAL 12 MONTH)
+    logger = Logger.new(STDOUT)
+    logger.info('Starting daily old user deletion')
+
+    total = 0
+    loop do
+      deleted_rows = DB[:sessions].with_sql_delete("
```
phantas:
```suggestion
      deleted_rows = DB[:userdetails].with_sql_delete("
```

PR #356, opened 2020-07-22
BAU: Use rubocop for linting

alan-gds:
The govuk-lint gem is deprecated; this PR switches us over to rubocop-govuk for linting, and then fixes the resulting failures.


.rubocop.yml
```diff
@@ -1,3 +1,4 @@
-Lint/AmbiguousBlockAssociation:
-  Exclude:
-    - "spec/**/*"
```
alan-gds, line 3: D'oh, I didn't spot that I'd changed this.  It's the reason I had to make fixes in spec/unit/lib/gdpr/gateway/user_details_spec.rb.  Probably better that we avoid unnecessary exclusions though; I can't see an obvious reason not to just use the style that the linter prefers in this case.

PR #359, opened 2020-08-06
Publish monthly and daily completion rate stats

alan-gds:
This is a first attempt.  We don't know much about how the Performance Platform actually works (and we might end up using some other mechanism to consume the published stats).

I've adapted the existing tasks in publish_statistics, rather than adding new ones, since I feel it'd be confusing to have multiple 'monthly' tasks (ditto for 'daily').  I''m assuming this will work, but who knows.

Not sure whether/how we can go about adding test coverage for this work.

https://trello.com/c/rtF5XZpO/

### DIFF
--- a/lib/performance_platform/gateway/completion_rate.rb
+++ b/lib/performance_platform/gateway/completion_rate.rb
@@ -1,11 +1,12 @@
 class PerformancePlatform::Gateway::CompletionRate
-  def initialize(date: Date.today.to_s)
+  def initialize(date: Date.today.to_s, period: "week")
     @date = Date.parse(date)
+    @period = period
   end
 
   def fetch_stats
     {
-      period: "week",
+      period: @period,
       metric_name: "completion-rate",
       sms_registered: sms_registered.count,
       sms_logged_in: sms_logged_in.count,
@@ -25,15 +26,15 @@ private
   end
 
   def sms_registered
-    repository.self_sign.with_sms.week_before(date)
+    repository.self_sign.with_sms.send("#{@period}_before", date)
   end
 
   def email_registered
-    repository.self_sign.with_email.week_before(date)
+    repository.self_sign.with_email.send("#{@period}_before", date)
   end
 
   def sponsor_registered
-    repository.sponsored.week_before(date)
+    repository.sponsored.send("#{@period}_before", date)
   end
 
   def sms_logged_in
--- a/tasks/publish_statistics.rb
+++ b/tasks/publish_statistics.rb
@@ -1,41 +1,36 @@
 require "logger"
 logger = Logger.new(STDOUT)
 
-task :publish_daily_statistics, :date do |_, args|
-  args.with_defaults(date: Date.today.to_s)
-  logger.info("publishing daily statistics with #{args}")
-  performance_gateway = PerformancePlatform::Gateway::PerformanceReport.new
-  volumetrics_gateway = PerformancePlatform::Gateway::Volumetrics.new(date: args[:date])
-  volumetrics_presenter = PerformancePlatform::Presenter::Volumetrics.new(date: args[:date])
+PERIODS = {
+  daily: "day",
+  weekly: "week",
+  monthly: "month",
+}.freeze
 
-  PerformancePlatform::UseCase::SendPerformanceReport.new(
-    stats_gateway: volumetrics_gateway,
-    performance_gateway: performance_gateway,
-  ).execute(presenter: volumetrics_presenter)
-end
+PERIODS.each do |adverbial, period|
+  name = "publish_#{adverbial}_statistics".to_sym
 
-task :publish_monthly_statistics, :date do |_, args|
-  args.with_defaults(date: Date.today.to_s)
-  logger.info("publishing monthly statistics with #{args}")
-  performance_gateway = PerformancePlatform::Gateway::PerformanceReport.new
-  volumetrics_gateway = PerformancePlatform::Gateway::Volumetrics.new(date: args[:date], period: "month")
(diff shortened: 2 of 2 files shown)


lib/performance_platform/gateway/completion_rate.rb
```diff
@@ -25,15 +26,36 @@ def repository
   end
 
   def sms_registered
-    repository.self_sign.with_sms.week_before(date)
+    case @period
+    when "day"
+      repository.self_sign.with_sms.day_before(date)
```
freesteph, line 31: you can tidy it up with 

```rb
repository.self_sign.with_sms.send("#{@period}_before", date)
```

lib/performance_platform/gateway/completion_rate.rb
```diff
@@ -25,15 +26,36 @@ def repository
...
   def sponsor_registered
-    repository.sponsored.week_before(date)
+    case @period
+    when "day"
+      repository.sponsored.day_before(date)
+    when "week"
+      repository.sponsored.week_before(date)
+    when "month"
+      repository.sponsored.month_before(date)
+    end
```
freesteph, line 58: I would advise to reduce all these cases with the dynamic method calling suggested above.

tasks/publish_statistics.rb
```diff
@@ -7,11 +7,18 @@
   performance_gateway = PerformancePlatform::Gateway::PerformanceReport.new
   volumetrics_gateway = PerformancePlatform::Gateway::Volumetrics.new(date: args[:date])
   volumetrics_presenter = PerformancePlatform::Presenter::Volumetrics.new(date: args[:date])
+  completion_rate_gateway = PerformancePlatform::Gateway::CompletionRate.new(date: args[:date], period: "day")
+  completion_rate_presenter = PerformancePlatform::Presenter::CompletionRate.new(date: args[:date])
```
freesteph, line 11: shameless plug to my way of [not duplicating these tasks](https://github.com/alphagov/govwifi-logging-api/blob/master/tasks/publish_statistics.rb#L13). Otherwise I'm worried it might get hard to read quite soon.

tasks/publish_statistics.rb
```diff
@@ -1,41 +1,36 @@
...
 
-  PerformancePlatform::UseCase::SendPerformanceReport.new(
-    stats_gateway: completion_rate_gateway,
-    performance_gateway: performance_gateway,
-  ).execute(presenter: completion_rate_presenter)
+    PerformancePlatform::UseCase::SendPerformanceReport.new(
+      stats_gateway: completion_rate_gateway,
+      performance_gateway: performance_gateway,
+    ).execute(presenter: completion_rate_presenter)
+  end
```
robinmitra, line 35: Shouldn't these be in `else` block?

alan-gds: No, the idea of this is: the completion rate stats should be sent regardless of what the period is (since we now want them daily, weekly and monthly), but the volumetrics stats should only be sent daily and monthly - i.e. not weekly.  The if block is there to exclude the volumetrics when the period is 'weekly'.  Happy to change it if anyone can suggest a clearer way to structure it.

tasks/publish_statistics.rb
```diff
@@ -1,41 +1,36 @@
...
+      ).execute(presenter: volumetrics_presenter)
+    end
 
-task :publish_weekly_statistics, :date do |_, args|
-  args.with_defaults(date: Date.today.to_s)
-  logger.info("publishing weekly statistics #{args}")
-  performance_gateway = PerformancePlatform::Gateway::PerformanceReport.new
-  completion_rate_gateway = PerformancePlatform::Gateway::CompletionRate.new(date: args[:date])
-  completion_rate_presenter = PerformancePlatform::Presenter::CompletionRate.new(date: args[:date])
+    completion_rate_gateway = PerformancePlatform::Gateway::CompletionRate.new(date: args[:date], period: "week")
```
robinmitra, line 28: As discussed, period should be parameterised to cover all periods.

tasks/publish_statistics.rb
```diff
@@ -1,41 +1,36 @@
...
+    performance_gateway = PerformancePlatform::Gateway::PerformanceReport.new
 
-  PerformancePlatform::UseCase::SendPerformanceReport.new(
-    stats_gateway: volumetrics_gateway,
-    performance_gateway: performance_gateway,
-  ).execute(presenter: volumetrics_presenter)
-end
+    if period != "week"
+      volumetrics_gateway = PerformancePlatform::Gateway::Volumetrics.new(date: args[:date])
+      volumetrics_presenter = PerformancePlatform::Presenter::Volumetrics.new(date: args[:date])
```
robinmitra, line 20: As discussed, period should be parameterised to cover all periods.

tasks/publish_statistics.rb
```diff
@@ -1,41 +1,36 @@
...
+    logger.info("Publishing #{adverbial} statistics with #{args[:date]}")
+    performance_gateway = PerformancePlatform::Gateway::PerformanceReport.new
 
-  PerformancePlatform::UseCase::SendPerformanceReport.new(
-    stats_gateway: volumetrics_gateway,
-    performance_gateway: performance_gateway,
-  ).execute(presenter: volumetrics_presenter)
-end
+    if period != "week"
+      volumetrics_gateway = PerformancePlatform::Gateway::Volumetrics.new(date: args[:date])
```
robinmitra, line 19: As discussed, period should be parameterised to cover all periods.

alan-gds:
Done (and thanks for explaining how to run it); it seems to be fine, once I throw enough environment variables at it.  Now that it's reviewed, I've squashed the commits together; please could someone add an approval.

PR #371, opened 2020-08-27
Add secrets baseline.

koetsier:
https://github.com/alphagov/gds-pre-commit

So we don't accidentally commit secrets

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,95 @@
+{
+  "exclude": {
+    "files": null,
+    "lines": null
+  },
+  "generated_at": "2020-08-26T19:27:33Z",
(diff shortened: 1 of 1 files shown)


robinmitra:
I couldn't test this as I couldn't get the install script for `gds-pre-commit` to run correctly (could be related to `pyenv`). Will try it again next week 🤞🏼

alan-gds (approved):
Looks fine to me - it prevents me from committing if I stick a fake password in password field in user.rb, and does not prevent me from committing if I don't add anything that looks like a secret.  I also confirmed that all of the values listed in 'results' are not real secrets, and are therefore fine to exist in the repository.

PR #379, opened 2020-09-10
Upload Volumetrics to S3

koetsier:
Data points about volumetrics that are normally sent to the PP
are now also sent to an S3 bucket. This because the Performance Platform is soon to be retired.

https://trello.com/c/ezleXA51/409-upload-volumetrics-data-points-to-s3


alan-gds (approved):
As far as I can tell, this all looks fine.

PR #381, opened 2020-09-11
Upload Completion Rate metrics to S3

koetsier:
Upload Completion Rate metrics to S3

These metrics are currently uploaded to the Performance Platform which will shortly be retired and so in order to preserve these datapoints we upload them to S3 instead.

### DIFF
--- a/lib/metrics/completion_rate.rb
+++ b/lib/metrics/completion_rate.rb
@@ -1,0 +1,30 @@
+module Metrics
+  class CompletionRate
+    VALID_PERIODS = %w[week day month].freeze
+
+    def initialize(attrs)
+      raise ArgumentError unless VALID_PERIODS.include? attrs[:period]
+
+      @period = attrs[:period]
+      @date = attrs[:date]
+    end
+
+    def execute
+      S3Publisher.publish key, stats
+    end
+
+    def key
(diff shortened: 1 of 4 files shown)


spec/unit/lib/metrics/volumetrics_spec.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+require_relative "./s3_fake_client"
+
+describe Metrics::Volumetrics do
+  let(:today) { Date.today }
+  let(:earlier_today) { today - 0.5 }
```
alan-gds, line 7: Won't this end up being yesterday, if it's called during the morning?  If we just mean "within the last 24 hours" though (as opposed to midnight to midnight), that's fine.

Same question applies for spec/unit/lib/metrics/completion_rate_spec.rb

alan-gds (commented):
Generally looks okay to me, I just have one question about some test setup.

spec/unit/lib/metrics/volumetrics_spec.rb
koetsier: Ah yes I see what you mean and that would be correct. I meant this to mean half a day ago. I'll see if I can rename this.

alan-gds (approved):
Looks fine to me.

PR #387, opened 2020-09-21
Remove rake tasks and code to send stats to the performance platform

koetsier:
The performance platform has been deprecated and so the code that sends
statistics to the Performance Platform can be removed.

Shared code has been moved to the Metrics namespace

Trello: https://trello.com/c/snhA5peo/411-remove-code-sending-data-to-the-performance-platform-only-after-data-s3-cards


robinmitra (approved):
LGTM

PR #403, opened 2020-10-22
chore: upgrade to Ruby 2.7.2

freesteph:
* upgrade to Bundle 2.x;
* upgrade CI alpine image;
* :warning: fix a bug in the launch sequence.


Gemfile.lock
```diff
@@ -159,7 +159,7 @@ DEPENDENCIES
   webmock
 
 RUBY VERSION
-   ruby 2.6.3p62
+   ruby 2.7.2p137
 
 BUNDLED WITH
-   1.17.2
+   2.1.4
```
alan-gds, line 165: Better to just get rid of BUNDLED WITH completely, right?

freesteph: I'm torn; on one hand, we could be liberal about it and say everyone bring your own `bundler`, but that will eventually bite us. I initially removed the line, but Bundler put it back in so I'd suggest we pin the version (i.e keep the line) but make sure it's 2.x.

robinmitra (approved):
🚀

PR #415, opened 2020-11-09
Fix/spec folder

freesteph:
Guard is very useful for TDD but it can't trigger if the spec files don't match the actual code. Remove the extra layer of indirection in the spec folder (`unit`).


Guardfile
```diff
@@ -0,0 +1,70 @@
...
+      rspec.spec.call("routing/#{m[1]}_routing"),
+      rspec.spec.call("controllers/#{m[1]}_controller"),
+      rspec.spec.call("acceptance/#{m[1]}"),
+    ]
+  end
+
+  # Rails config changes
+  watch(rails.spec_helper)     { rspec.spec_dir }
+  watch(rails.routes)          { "#{rspec.spec_dir}/routing" }
+  watch(rails.app_controller)  { "#{rspec.spec_dir}/controllers" }
```
robinmitra, line 59: Can probably get rid of this.

Guardfile
```diff
@@ -0,0 +1,70 @@
...
+
+  # Capybara features specs
+  watch(rails.view_dirs)     { |m| rspec.spec.call("features/#{m[1]}") }
+  watch(rails.layouts)       { |m| rspec.spec.call("features/#{m[1]}") }
```
robinmitra, line 63: ...and these.

robinmitra:
Looks good, but would be nice to get rid of the Rails crufts from the `Guardfile`.

robinmitra (approved):

PR #420, opened 2020-11-13
Reset RACK_ENV in specs where it's overridden

colinbm:
In some tests, `ENV["RACK_ENV"]` is overridden to `staging` or `production`
either in a `before` block or inline.

ENV is not reset between tests, so any subsequent tests requiring `RACK_ENV`
to be `test` will then fail.

This change moves inlines to `before` blocks, and adds an `after` block
to reset back to `test`.

This has come up as part of the survey work, however it seems worth isolating in a separate PR.

### DIFF
--- a/spec/acceptance/sms_notification_spec.rb
+++ b/spec/acceptance/sms_notification_spec.rb
@@ -45,18 +45,30 @@ RSpec.describe App do
       end
 
       context "production" do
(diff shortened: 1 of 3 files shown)


freesteph (approved):
Great find, thanks. That explains our intermittent failures on the survey branch.

PR #424, opened 2020-11-23
fix: fix the factory and timing logic for the inactive users survey

freesteph:
Couple of issues here:

* there was a `focus: true` in the spec;
* the created_at logic should be 14 days ago inclusive (from midnight,
i.e created_at >=) all the way *before* 13 days ago (i.e created_at
<) rather than the opposite. Effectively that's a one-second interval
we're shifting but it's semantically correct, and fails the test suite
otherwise.


robinmitra (approved):
🕶️

PR #426, opened 2020-11-25
Change survey log message to say "survey" not "email"

colinbm:
Surveys are sent via email, and SMS, and we don't know how many of which
when we log this.


freesteph (approved):

freesteph (changes requested):
Yes to the work, no to the branch :smile:  it's out of sync with master and the branch name will be confusing in the merge commit.

Can you either rename the branch or open a new one rebased off master?

PR #446, opened 2021-01-22
Task to delete smoke test users

colinbm:
The new smoke-tests running regularly will create a new user each time which will create noise in the database, and skew user stats.

This task deletes any users with a contact email like `govwifi-tests+%@digital.cabinet-office.gov.uk`. It skips users created in the last 10 minutes, to avoid any potential conflict with the smoke test itself running.


lib/smoke_tests/gateway/user_details.rb
```diff
@@ -0,0 +1,26 @@
...
+        AND created_at < NOW() - INTERVAL 10 MINUTE
+        LIMIT #{SESSION_BATCH_SIZE}
+      ")
+      total += deleted_rows
+
+      if deleted_rows.zero?
+        break
+      end
+
+      logger.info("Finished daily smoke test user deletion, #{total} rows affected")
```
koetsier, line 23: should this not be outside the loop?

lib/smoke_tests/gateway/user_details.rb
```diff
@@ -0,0 +1,26 @@
+require "logger"
+
+class SmokeTests::Gateway::UserDetails
+  SESSION_BATCH_SIZE = 500
+  def delete_users
+    logger = Logger.new(STDOUT)
+    logger.info("Starting daily smoke test user deletion")
+
+    total = 0
+    loop do
```
koetsier, line 10: Would a `while` loop be possible here

lib/smoke_tests/gateway/user_details.rb
```diff
@@ -0,0 +1,26 @@
...
+      deleted_rows = DB[:userdetails].with_sql_delete("
```
koetsier, line 11: could we use the Sequel interface here and the ```user``` model defined in this repo?

spec/lib/smoke_tests/gateway/user_details_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    context "Given multiple test users" do
+      before do
+        user_details.insert(username: "foo", contact: "govwifi-tests+foo@digital.cabinet-office.gov.uk", created_at: Date.today - 1)
+        user_details.insert(username: "bar", contact: "govwifi-tests+bar@digital.cabinet-office.gov.uk", created_at: Date.today - 1)
+        user_details.insert(username: "baz", contact: "govwifi-tests+baz@digital.cabinet-office.gov.uk", created_at: Date.today - 1)
+      end
+
+      it "deletes all the test users" do
+        subject.delete_users
+        expect(user_details.all.map { |s| s.fetch(:username) }).to be_empty
```
koetsier, line 40: is the ```map``` necessary?

koetsier (changes requested):
A few comments but looks good 👍

lib/smoke_tests/gateway/user_details.rb
colinbm: :+1:

lib/smoke_tests/gateway/user_details.rb
colinbm: We could... this was created as a clone of https://github.com/alphagov/govwifi-user-signup-api/blob/master/lib/gdpr/gateway/usedetails.rb so mainly replicates the structure there - same for some other points.

spec/lib/smoke_tests/gateway/user_details_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    context "Given one test user" do
+      before do
+        user_details.insert(username: "foo", contact: "foo@example.com", created_at: Date.today - 1)
+        user_details.insert(username: "bar", contact: "bar@digital.cabinet-office.gov.uk", created_at: Date.today - 1)
+        user_details.insert(username: "baz", contact: "govwifi-tests+baz@digital.cabinet-office.gov.uk", created_at: Date.today - 1)
+      end
+
+      it "does deletes only the old user record" do
```
robinmitra, line 25: I think it should be this instead?
```suggestion
      it "does deletes only the test user record" do
```

robinmitra (approved):
LGTM

PR #454, opened 2021-02-09
Long term fix for SMS bot-loop issue

colinbm:
This reverts the quick fix from Friday which disabled the default message to break the loop.

The new solution follows these rules:

* If SAME message from same number received 3 times in 15 minutes, don't reply
* If ANY message from same number received 3 times in 5 minutes, don't reply

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -103,7 +103,9 @@ class App < Sinatra::Base
 
   def sender_is_repetitive?(source, message)
     contact_sanitiser = WifiUser::UseCase::ContactSanitiser.new
-    repetitive_sms_checker = WifiUser::UseCase::RepetitiveSmsChecker.new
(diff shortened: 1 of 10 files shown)


camdesgov:
This looks good to me, but I think it would be good to get another developer to check it over too. I've assigned Jos

app.rb
```diff
@@ -78,6 +78,11 @@ class App < Sinatra::Base
       return ""
     end
 
+    if sender_is_repetitive?(source, message)
+      logger.warn("Possible bot-number detected: #{source}")
```
koetsier, line 82: Maybe just say here that we received too many messages instead? At a later date when we've forgotten about the incident, we may not understand this

koetsier (approved):
Looks good to me. Maybe you could add a link to the trello card in the PR description for future reference.

PR #464, opened 2021-03-04
Handle Notifications::Client::BadRequestError

colinbm:
If a text message comes in from an invalid number, a `Notifications::Client::BadRequestError` is raised. There are also other circumstances we haven't seen that could trigger this, such as a text from a landline number.

Catch this error, and if it is due to a `ValidationError` (in the body of the message) log it and rollback. Else throw the original error.

Either way, we're now wrapping the user creation in a transaction, so on either error the user will not be created in the database.

### DIFF
--- a/lib/wifi_user/use_case/sms_response.rb
+++ b/lib/wifi_user/use_case/sms_response.rb
@@ -22,7 +22,6 @@ class WifiUser::UseCase::SmsResponse
     message = validation_errors.map { |err| err["message"] }
                                .join(", ")
     logger.warn("Failed to send email: #{message}")
-    raise Sequel::Rollback
   end
 
 private
--- a/lib/wifi_user/use_case/sms_response.rb
+++ b/lib/wifi_user/use_case/sms_response.rb
@@ -9,9 +9,20 @@ class WifiUser::UseCase::SmsResponse
     phone_number = WifiUser::UseCase::ContactSanitiser.new.execute(contact)
     return logger.warn("Unexpected contact detail found #{contact}") if phone_number.nil?
 
-    login_details = user_model.generate(contact: phone_number)
-    notify_params = { login: login_details[:username], pass: login_details[:password] }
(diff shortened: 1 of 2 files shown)


lib/wifi_user/use_case/sms_response.rb
```diff
@@ -9,9 +9,20 @@ def execute(contact:, sms_content:)
...
+      send_signup_instructions(phone_number, notify_params, sms_content)
+    end
+  rescue Notifications::Client::BadRequestError => e
+    validation_errors = e.message.select { |err| err["error"] == "ValidationError" }
+    raise e if validation_errors.empty?
+
+    message = validation_errors.map { |err| err["message"] }
+                               .join(", ")
+    logger.warn("Failed to send email: #{message}")
+    raise Sequel::Rollback
```
koetsier, line 25: The documentation is a bit vague about this, but is it necessary to raise a rollback here? If I read the docs correctly, on an exception, the transaction rolls back and re-raises the error which would get caught here. Can you leave it at that?

Another thing: if line 20 raises an exception then presumably the rollback wouldn't be raised?

The docs are a bit sparse so I could easily be really wrong here 😄

koetsier (changes requested):

lib/wifi_user/use_case/sms_response.rb
colinbm: At the point where we call `Sequel::Rollback` we've caught and logged the error, so no exception will be raised. That means the transaction will complete, and write the record to the database. As we won't have sent the details to the user, there's not much point in doing this. 

`Sequel::Rollback` is treated differently to any other exception - it triggers the transaction rollback, but *does not then get re-raised*, which means we won't see it in Sentry in that instance, and the code continues (in this case just responding with a `200` to the http request.

koetsier (approved):
lgtm! 👍

PR #484, opened 2021-05-10
[DO NOT MERGE] Adjust template IDs for Chromebook and "recap" messages 

colinbm:
## Note

Pushing this through to production needs to be orchestrated so that the final edit to the Notify template is done at the same time as deploy.

## What

Adds Chromebook (6) and Recap (default) templates.

## Why

Chromebook - adds missing instructions

Recap - previously the default message was the "generic device" instructions... however if you've failed to get to the "Go" point, then these won't help... the new message give useful information regardless of which stage you're stuck at.


koetsier (approved):
looks good!

PR #521, opened 2021-08-09
Remove the limiting code so 100% of users are sent the surveys

colinbm:
### What

Remove the limiting code that only sent the surveys to 25% of users

### Why

Because we want to send the surveys to everyone.

Link to Trello card (if applicable): https://trello.com/c/6Bwk5hRq


cbaines (approved):
Sounds good to me :+1:

PR #530, opened 2021-09-08
Don't record an smslog entry for messages with no number.

colinbm:
### What

If we get a text message without a number, don't record it in the smslog.

### Why

Sometimes we get text messages with no number. These are typically spam, or someone has entered our number somewhere by accident. These typically appear on phones as coming from a text based source, e.g. `GOOGLE`

As there's no actual number, it tries to insert a record with `nil`. The database is configured with number `NOT NULL` so that errors.

As we'd have no way of replying to these messages anyway, there's no need to record them for purposes of preventing a loop.

Link to Trello card (if applicable): https://trello.com/c/g0u55jLg


szd55gds (approved):
This looks ok to me

PR #534, opened 2021-09-20
Extend GDPR deletion threshold to 2 years

colinbm:
### What
Extend GDPR deletion threshold to 2 years

### Why
We're reaching 18 months and there's still people not back in the office


sarahseewhy (approved):
LGTM 👍

PR #559, opened 2022-01-07
The docker-compose file now uses the healthcheck feature

koetsier:
We now use the docker-compose healthcheck feature to make sure
that MySql is running before the app starts

The mysql databases are now started by defining dependencies
within the docker-compose file, instead of starting them up
separately in the makefile.


cbaines (approved):
I'd prefer the commit message being written in the imperative, so "Use healthcheck feature in docker-compose file", but the changes look fine to me :+1:

PR #569, opened 2022-02-03
let's set a sensible name rather than the filename

szd55gds:
### What
Add a 'name' for our lint & test pipeline

### Why
The default is purely the filename and this does not describe the function of this workflow. We may add further workflows, so let's label this sensibly.


sarahseewhy (approved):
🚀

PR #571, opened 2022-02-07
Bump `notifications-ruby-client` gem

sarahseewhy:
### What

* Bump `notifications-ruby-client gem` to v.5.0.3 (resolves PR #511 )
* Accommodate updates from `notifications-ruby-client gem` v.5.0.3 by replacing array methods (e.g., `dig`, `select`) with string methods (`.include?`).
* Small stylistic refactoring (use `unless` instead of `if not`, remove unnecessary `validation_errors` variable declaration). 

Thank you to @cbaines who helped review the Ruby code and made some very helpful suggestions.

### Why

Bumping to a later version of `notifications-ruby-client` means we have to update our code. In the v.4.0 release notes for the gem it states: 

> RequestError.message now returns a string, not an array of hashes – see alphagov/notifications-ruby-client#72

This means we must change how we interrogate `error.message` since it now returns a string not an array of hashes.

Once this PR is merged I will close the dependabot PR #511 

Link to Trello card (if applicable): https://trello.com/c/QPe3jqwC/1995-lets-track-dependabots-so-we-have-evidence-of-this-work-and-a-record-of-the-ones-weve-completed


cbaines (approved):
Looks good to me :+1:

PR #573, opened 2022-02-10
Handle parse errors when parsing email addresses for user signup

cbaines:
### What
Handle parse errors when parsing email addresses for user signup

### Why
So that these are just logged, rather than making it to Sentry.

Link to Trello card: https://trello.com/c/lzTo5ku0/1998-dont-alert-sentry-on-malformed-email-address


sarahseewhy (approved):
This looks fine to me, though I'm not a confident Rubyist.

PR #598, opened 2022-05-06
Remove the inactive user survey code

koetsier:
### What
Remove the inactive user survey code

### Why
The inactive user survey is no longer needed.

See also: [remove_inactive_user_survey](https://github.com/alphagov/govwifi-user-signup-api/pull/573)

https://trello.com/c/QwI73D5c/2215-close-2020-inactive-user-survey-api-integration-with-notify


cbaines (approved):
Sounds OK to me, providing we're not going to send out this survey in the future?

PR #600, opened 2022-05-17
Reply to rejected email signups

cbaines:
### What
Reply to rejected email signups

### Why
The intent here is to let people know they won't get credentials when
they try to signup with an unrecognised email address, and direct them
to either signup via text message or have someone sponsor them.

Link to Trello card: https://trello.com/c/84mUFy3f/2266-let-unrecognised-email-domains-know-they-wont-get-credentials

### DIFF
--- a/spec/lib/wifi_user/use_cases/email_signup_spec.rb
+++ b/spec/lib/wifi_user/use_cases/email_signup_spec.rb
@@ -89,9 +89,18 @@ describe WifiUser::UseCase::EmailSignup do
 
       context "given an email address with a non-gov domain" do
         let(:whitelist_checker) { double(execute: { success: false }) }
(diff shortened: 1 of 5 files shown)


spec/lib/wifi_user/use_cases/email_signup_spec.rb
```diff
@@ -100,7 +109,7 @@
         end
 
         it "does not send an email to Notify" do
```
koetsier, line 111: This test should probably be renamed say that we now do send a message

koetsier, line 111: This test should probably be renamed to show that we do send a message to Notify now

cbaines: Good spot, I've renamed the test.

PR #604, opened 2022-06-16
Omit hash value where possible

koetsier:
### What
Omit hash value where possible
### Why
This so that the linter still passes when we upgrade


koetsier:
duplicate

PR #637, opened 2022-10-21
Revert "Introduce an AWS Codebuild Task"

camdesgov:
Reverts alphagov/govwifi-user-signup-api#634

Merged in error reverting.


szd55gds (approved):
Approved revert

PR #657, opened 2022-12-14
GW-647 Send reminder of credentials to Sponsored Users when they email 'sign…

abdullaha1985:
…up@wifi.service.gov.uk'

Sponsored Users could not get reminder of credentials when they email 'signup@wifi.service.gov.uk'. This commit adds a check to see if the Sponsee's Sponsor email address is valid, If so, we will send the normal email reminder of credentials normal.

### What
Describe the change
Given a Sponsee sends a blank email to 'signup@wifi.service.gov.uk'.
This PR checks the Sponsee's email address exists and their Sponsor email is valid.
If so, then the Sponsee will receive normal credentials reminder

### Why
Describe why the change was necessary
This fix is necessary because Sponsored Users were told they should email 'signup@wifi.service.gov.uk'
in-order to get reminder of their credentials, but Sponsee User receives not response.

https://technologyprogramme.atlassian.net/browse/GW-647

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -38,9 +38,14 @@ class App < Sinatra::Base
       ),
     )
 
+    check_user_is_sponsee = WifiUser::UseCase::CheckUserIsSponsee.new(
+      allowlist_checker:,
+    )
+
     email_signup_handler = ::WifiUser::UseCase::EmailSignup.new(
       user_model: WifiUser::Repository::User.new,
       allowlist_checker:,
+      check_user_is_sponsee:,
       logger:,
     )
 
--- a/app.rb
+++ b/app.rb
@@ -38,14 +38,12 @@ class App < Sinatra::Base
       ),
     )
 
-    check_user_is_sponsee = WifiUser::UseCase::CheckUserIsSponsee.new(
-      allowlist_checker:,
-    )
+    sponsee_is_user_checker = WifiUser::UseCase::CheckUserIsSponsee.new
 
     email_signup_handler = ::WifiUser::UseCase::EmailSignup.new(
       user_model: WifiUser::Repository::User.new,
       allowlist_checker:,
-      check_user_is_sponsee:,
+      sponsee_is_user_checker:,
       logger:,
     )
 
--- a/lib/wifi_user/use_case/check_user_is_sponsee.rb
+++ b/lib/wifi_user/use_case/check_user_is_sponsee.rb
@@ -1,20 +1,7 @@
 class WifiUser::UseCase::CheckUserIsSponsee
-  def initialize(allowlist_checker:)
-    @allowlist_checker = allowlist_checker
-  end
-
   def execute(contact)
-    sponsee = DB[:userdetails].where(contact:)&.first
+    sponsee = WifiUser::Repository::User.find(contact:)
 
-    return false unless sponsee
-    return false unless sponsee[:sponsor]
-
-    sponsor_email = sponsee[:sponsor]
-
-    allowlist_checker.execute(sponsor_email)[:success]
+    sponsee.present? && sponsee[:sponsor] != contact
   end
-
-private
-
-  attr_reader :allowlist_checker
 end
--- a/lib/wifi_user/use_case/check_user_is_sponsee.rb
+++ b/lib/wifi_user/use_case/check_user_is_sponsee.rb
@@ -1,0 +1,20 @@
+class WifiUser::UseCase::CheckUserIsSponsee
+  def initialize(allowlist_checker:)
+    @allowlist_checker = allowlist_checker
+  end
+
+  def execute(contact)
+    sponsee = DB[:userdetails].where(contact:)&.first
+
+    return false unless sponsee
+    return false unless sponsee[:sponsor]
+
+    sponsor_email = sponsee[:sponsor]
+
+    allowlist_checker.execute(sponsor_email)[:success]
+  end
+
+private
+
+  attr_reader :allowlist_checker
+end
--- a/lib/wifi_user/use_case/email_signup.rb
+++ b/lib/wifi_user/use_case/email_signup.rb
@@ -2,18 +2,20 @@ require "mail"
 require "notifications/client"
 
 class WifiUser::UseCase::EmailSignup
-  def initialize(user_model:, allowlist_checker:, check_user_is_sponsee:, logger: Logger.new($stdout))
+  def initialize(user_model:, allowlist_checker:, sponsee_is_user_checker:, logger: Logger.new($stdout))
     @user_model = user_model
     @allowlist_checker = allowlist_checker
-    @check_user_is_sponsee = check_user_is_sponsee
+    @sponsee_is_user_checker = sponsee_is_user_checker
     @logger = logger
   end
 
   def execute(contact:)
     email_address = Mail::Address.new(contact).address
 
-    if allowlist_checker.execute(email_address)[:success] || check_user_is_sponsee.execute(email_address)
+    if allowlist_checker.execute(email_address)[:success]
       send_signup_instructions(email_address)
+    elsif sponsee_is_user_checker.execute(email_address)
+      send_sponsor_signup_instructions(email_address)
     else
       logger.info("Unsuccessful email signup attempt: #{email_address}")
       send_rejected_email_address_email(email_address)
@@ -24,7 +26,7 @@ class WifiUser::UseCase::EmailSignup
 
 private
 
-  attr_accessor :user_model, :allowlist_checker, :logger, :check_user_is_sponsee
+  attr_accessor :user_model, :allowlist_checker, :logger, :sponsee_is_user_checker
 
   def send_signup_instructions(email_address)
     client = Notifications::Client.new(ENV.fetch("NOTIFY_API_KEY"))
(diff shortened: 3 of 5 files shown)


lib/wifi_user/use_case/check_user_is_sponsee.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def execute(contact)
+    sponsee = DB[:userdetails].where(contact:)&.first
+
+    return false unless sponsee
+    return false unless sponsee[:sponsor]
+
+    sponsor_email = sponsee[:sponsor]
+
+    allowlist_checker.execute(sponsor_email)[:success]
```
koetsier, line 14: This is probably not necessary? If the user has been saved correctly, it should have passed the allowlist already right?

lib/wifi_user/use_case/check_user_is_sponsee.rb
```diff
@@ -0,0 +1,20 @@
+class WifiUser::UseCase::CheckUserIsSponsee
+  def initialize(allowlist_checker:)
+    @allowlist_checker = allowlist_checker
+  end
+
+  def execute(contact)
+    sponsee = DB[:userdetails].where(contact:)&.first
+
+    return false unless sponsee
+    return false unless sponsee[:sponsor]
```
koetsier, line 10: A user is sponsored if the contact field is not the same as the sponsor field. The sponsor field is always populated (https://github.com/alphagov/govwifi-user-signup-api/blob/7c45f2c3c60f91f1032b1de695f0a62e4d0df204/lib/wifi_user/repository/user.rb#L6 - when the sponsor is nil, it is the same as 'contact'

lib/wifi_user/use_case/check_user_is_sponsee.rb
```diff
@@ -0,0 +1,20 @@
+class WifiUser::UseCase::CheckUserIsSponsee
+  def initialize(allowlist_checker:)
+    @allowlist_checker = allowlist_checker
+  end
+
+  def execute(contact)
+    sponsee = DB[:userdetails].where(contact:)&.first
```
koetsier, line 7: Can use find_by_contact here

lib/wifi_user/use_case/check_user_is_sponsee.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    return false unless sponsee
```
koetsier, line 9: better combine these 2 statements

lib/wifi_user/use_case/email_signup.rb
```diff
@@ -2,16 +2,17 @@
...
     @allowlist_checker = allowlist_checker
+    @check_user_is_sponsee = check_user_is_sponsee
     @logger = logger
   end
 
   def execute(contact:)
     email_address = Mail::Address.new(contact).address
 
-    if allowlist_checker.execute(email_address)[:success]
+    if allowlist_checker.execute(email_address)[:success] || check_user_is_sponsee.execute(email_address)
```
koetsier, line 15: The main problem here is that this will render the wrong template - it sends the standard user signup message and it probably should send the sponsor signup one

app.rb
```diff
@@ -38,9 +38,14 @@ class App < Sinatra::Base
...
     )
 
+    check_user_is_sponsee = WifiUser::UseCase::CheckUserIsSponsee.new(
+      allowlist_checker:,
+    )
+
     email_signup_handler = ::WifiUser::UseCase::EmailSignup.new(
       user_model: WifiUser::Repository::User.new,
       allowlist_checker:,
+      check_user_is_sponsee:,
```
koetsier, line 48: this should probably be named `sponsee_is_user_checker` to keep it consistent

koetsier (changes requested):
I think you forgot the SMS journey for sponsored users as well. Sponsored users should be able to send an sms and get their credentials back.  (https://github.com/alphagov/govwifi-user-signup-api/blob/7c45f2c3c60f91f1032b1de695f0a62e4d0df204/app.rb#L67)

lib/wifi_user/use_case/check_user_is_sponsee.rb
abdullaha1985: Thats correct **allowlist_checker** is used when creating User
 
My thinking was the database might not always have correct data.

abdullaha1985: > I think you forgot the SMS journey for sponsored users as well. Sponsored users should be able to send an sms and get their credentials back. (
> 
> https://github.com/alphagov/govwifi-user-signup-api/blob/7c45f2c3c60f91f1032b1de695f0a62e4d0df204/app.rb#L67
> 
> )

The sms journey looks the same for Sponsees and normal User e.g. Both use the same Templates.

spec/lib/wifi_user/use_cases/email_signup_spec.rb
```diff
@@ -117,10 +119,25 @@
 
       context "given an email address with a non-gov domain, but email is sponsored" do
         let(:allowlist_checker) { double(execute: { success: false }) }
-        let(:check_user_is_sponsee) { double(execute: true) }
-        let(:created_contact) { "ryan@example.com" }
-        let(:username) { "irrelevant" }
-        let(:password) { "irrelephant" }
+        let(:sponsee_is_user_checker) { double(execute: true) }
```
koetsier, line 122: I don't think any of these, apart from maybe created_contact require `let!` do they? Better use plain `let` I think

spec/lib/wifi_user/use_cases/email_signup_spec.rb
```diff
@@ -112,6 +116,39 @@
...
+            personalisation: {
+              username:,
+              password:,
+              sponsor:,
+            },
+            email_reply_to_id: do_not_reply_id,
+          }
+        end
+
+        it "returns credentials username and password" do
```
koetsier, line 142: Not sure if we are bothered whether `user_model` receives `generate` or not? The only thing this is supposed to do is send an email right?

lib/wifi_user/use_case/email_signup.rb
```diff
@@ -46,6 +50,18 @@ def send_rejected_email_address_email(email_address)
     )
   end
 
+  def send_sponsor_signup_instructions(email_address)
+    sponsee = WifiUser::Repository::User.find(contact: email_address)
+    login_details = user_model.generate(contact: email_address)
```
koetsier, line 55: maybe better use `find` here, as we're not going to create a user

lib/wifi_user/use_case/email_signup.rb
```diff
@@ -46,6 +50,18 @@ def send_rejected_email_address_email(email_address)
...
+
+    send_email_gateway.execute(
```
koetsier, line 57: line 43 and line 32 don't inject an email gateway. Better stay consistent and it removes one more parameter / injection in the constructor

spec/lib/wifi_user/use_cases/check_user_is_sponsee_spec.rb
```diff
@@ -0,0 +1,32 @@
+describe WifiUser::UseCase::CheckUserIsSponsee do
+  subject { described_class.new }
+
+  let!(:normal_user) { FactoryBot.create(:user_details, :recent, :active, :self_signed) }
+  let!(:sponsored_user) { FactoryBot.create(:user_details, :recent, :active, :sponsored) }
```
koetsier, line 5: this line doesn't require `let!` Maybe even better put these lines in the 'before` block

spec/lib/wifi_user/use_cases/check_user_is_sponsee_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  before do
+    @contact = sponsored_user.contact
+    @sponsor = sponsored_user.sponsor
+  end
+
+  context "given user is sponsored" do
+    it "returns true when sponsee has a valid sponsor email" do
+      result = subject.execute(sponsored_user.contact)
```
koetsier, line 14: you define `@sponsor` already in the `before` block

spec/lib/wifi_user/use_cases/check_user_is_sponsee_spec.rb
```diff
@@ -0,0 +1,32 @@
+describe WifiUser::UseCase::CheckUserIsSponsee do
+  subject { described_class.new }
+
+  let!(:normal_user) { FactoryBot.create(:user_details, :recent, :active, :self_signed) }
+  let!(:sponsored_user) { FactoryBot.create(:user_details, :recent, :active, :sponsored) }
+
+  before do
+    @contact = sponsored_user.contact
```
koetsier, line 8: are you using these?

koetsier (changes requested):
Please sort out those commits!

And this PR hasn't addressed the SMS route yet. Sponsored users can also send an SMS which needs a reply sent. This uses the `/user-signup/sms-notification/notify` route in `app.rb`

spec/lib/wifi_user/use_cases/email_signup_spec.rb
abdullaha1985: On line 109 we are testing`expect(user_model).not_to receive(:generate)` given User is non-gov.
I assumed we want to test the opposite if User is Sponsored?

abdullaha1985:
The journey for when Sponsee User sends a SMS for credentials they currently get a response from the system.
I believe that it's irrelevant if they're sponsored or not. So if they signed up themselves or someone else signed them up, if they text us, we will reply with details.

spec/lib/wifi_user/use_cases/email_signup_spec.rb
koetsier: This test is supposed to test that an email is sent when a sponsored user sends a message right? The `User.generate` method is basically a `find_or_create` call and I think it may be a bit of an implementation detail if there is code present that finds a user. If you want to test that no additional user is being created it may be easier to do

```
expect {
  subject.execute(contact: created_contact)
}.to_not change { User.count }
```

koetsier:
The user signup api has changed a lot since this ticket, so merging is going to be very hard. There are also many things wrong with this PR so I'll close it

PR #664, opened 2023-01-04
Add extra feature specs

koetsier:
### What
Add extra feature specs
### Why
The code in this repo uses a lot of mocking to test components, which can make them brittle

This PR adds some feature level tests that use minimal mocking to test the functionality of
the application


iram-shehzadi (approved):
Looks good to me 👍

PR #665, opened 2023-01-10
Refactor email signup and sponsor signup journeys

koetsier:
### What
Refactor email signup and sponsor signup journeys
### Why
The code before was hard to read and relied on excessive mocking to test the code, making tests brittle and unreliable.

The resulting code approximately 900 lines shorter


A1Robson:
Seems to be a makefile bug from previous removal of concourse. 
line 1 of makefile should possibly be: DOCKER_COMPOSE = docker-compose -f docker-compose.yml

A1Robson:
Also, the test isn't actually working on this, not sure why its reporting success in the checks above - have a look at the logs under the test action: [https://github.com/alphagov/govwifi-user-signup-api/actions/runs/5530213407/job/14975180958?pr=665](url)

A1Robson (approved):
lgtm.
wonder if its worth creating a package to share some common objects between this code and admin site ?

koetsier:
> Seems to be a makefile bug from previous removal of concourse. line 1 of makefile should possibly be: DOCKER_COMPOSE = docker-compose -f docker-compose.yml

I rebased the code so this should be fixed

koetsier:
> Also, the test isn't actually working on this, not sure why its reporting success in the checks above - have a look at the logs under the test action: [https://github.com/alphagov/govwifi-user-signup-api/actions/runs/5530213407/job/14975180958?pr=665](url)

I looked into this and, because of the docker-compose error the '-f docker-compose' command is  run within the makefile - which is erroneous, it should run the docker-compose command. The makefile actually interprets the leading minus in the '-f' command as as a directive to not stop processing the makefile, even if an error is encountered. So the makefile keeps going until the last command which is 'rm -rf' which succeeds and returns error code 0.

PR #689, opened 2023-03-21
Ensure the s3 region is correct and consistent

koetsier:
### What
The S3 client was always instantiated with eu-west-2, but sometimes the resource was instantiated with eu-west-1. This commit makes these two the same value
### Why
This causes confusion and is the source of a possible issue


szd55gds (approved):
LGTM

PR #723, opened 2023-08-29
Fix docker compose

A1Robson:
### What
Makefile was missing a DOCKER_COMPOSE variable
Update platform so that works on m1 Mac

### Why
enable local development


A1Robson:
work already been done. not needed

PR #724, opened 2023-09-05
Require 'yaml' library to fix issue in active user survey

koetsier:
The active user survey fails because it hasn't loaded the yaml library.

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-1116


koetsier:
part of new_refactor branch

PR #795, opened 2024-06-10
Add rake task and logic to send a followup message

koetsier:
### What
Add rake task and logic to send a followup message

When a user has not logged in for more than 2 days after signing up
we want to send a message to help the user and ask why they did not manage to
connect.
### Why
We want to increase uptake of GovWifi and to identify and help people with technical problems to connect.
We want to figure out why some people ask for credentials but then don't connect

https://technologyprogramme.atlassian.net/browse/GW-1689


leonaAtkins:
You've not added a config to Alpaca, will this cause the smoke tests to fail?

koetsier:
> You've not added a config to Alpaca, will this cause the smoke tests to fail?

It's a rake task that only really makes sense in production because it sends messages to users. As long as we don't call the rake task (using AWS eventbridge) it shouldn't cause any issues.

leonaAtkins (approved):

PR #811, opened 2024-07-03
add notifications for inactive and deleted users

iram-shehzadi:
### What
Send notification before user credentials are due to expire - 11 months
Send notification when user credentials have expired and user is removed with automated rake task - 12 months

### Why
We get support requests from users regarding accounts that are not working and this is due to account expiring therefor it it would be beneficial for us to notify users before and after credentials expire.

Link to Jira card (if applicable): 
https://technologyprogramme.atlassian.net/browse/GW-167
https://technologyprogramme.atlassian.net/browse/GW-1760

### DIFF
--- a/spec/lib/gdpr/gateway/user_details_spec.rb
+++ b/spec/lib/gdpr/gateway/user_details_spec.rb
@@ -196,7 +196,7 @@ describe Gdpr::Gateway::Userdetails do
       expect(notify_client).to receive(:send_sms).with(
         phone_number: "+07391491234",
         template_id: "3d24dcb3-503c-4045-9ac4-22d0868eb49f",
-        personalisation: { inactivity_period: "11 months", username: "Rob" }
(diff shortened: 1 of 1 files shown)


lib/gdpr/gateway/user_details.rb
```diff
@@ -29,4 +55,32 @@ def obfuscate_sponsors
...
+      "),
+    ).all
+
+    logger.info("Found #{inactive_users.size} inactive users")
+
+    inactive_users.each do |user|
+      contact = user[:contact]
+      username = user[:username]
+
+      if contact.start_with?("+")
```
koetsier, line 77: you could move this to the WifiUser model, so you could use `if contact.email?` so this logic doesn't need to be duplicated

koetsier (approved):
Looks good to me. I made one small suggestion.

PR #816, opened 2024-07-18
error handling

iram-shehzadi:
### What
Errors due to no error handling

### Why
Handle errors so the rake runs

### DIFF
--- a/lib/gdpr/gateway/user_details.rb
+++ b/lib/gdpr/gateway/user_details.rb
@@ -64,7 +64,7 @@ class Gdpr::Gateway::Userdetails
 
   def notify_inactive_users
     @logger.info("Starting notification process for users inactive for 11 months")
-    inactive_users = find_users_inactive_for_11_months(DAYS_IN_11_MONTHS)
+    inactive_users = find_inactive_users_for_exact_number_of_days(DAYS_IN_11_MONTHS)
 
     @logger.info("Found #{inactive_users.size} inactive users")
 
@@ -102,7 +102,7 @@ private
     ).all
   end
 
-  def find_users_inactive_for_11_months(days)
+  def find_inactive_users_for_exact_number_of_days(days)
     WifiUser::User.where(
       Sequel.lit("
       (DATE(last_login) = DATE_SUB(CURDATE(), INTERVAL ? DAY)


lib/gdpr/gateway/user_details.rb
```diff
@@ -2,37 +2,37 @@
...
       if user.mobile?
         WifiUser::SMSSender.notify_user(username, contact)
-        logger.info("SMS sent to #{username}: #{contact}")
       elsif user.valid_email?(contact)
-        WifiUser::EmailSender.notify_user(username, contact)
-        logger.info("Email sent to #{username}: #{contact}")
+        begin
+          WifiUser::EmailSender.notify_user(username, contact)
+        rescue Notifications::Client::BadRequestError => e
+          handle_email_error(e, username, contact)
```
koetsier, line 28: we may want to rescue StandardError as well here so we're certain the delete cycle runs

lib/gdpr/gateway/user_details.rb
```diff
@@ -79,12 +71,36 @@ def notify_inactive_users
...
       end
     end
 
-    logger.info("Finished notification process for users who were inactive for 11 months")
+    @logger.info("Finished notification process for users inactive for 11 months")
+  end
+
+private
+
+  def find_inactive_users(months)
```
koetsier, line 90: Does this cause inactive users being sent this email every day for a month? (assuming this runs daily)

koetsier (changes requested):

lib/gdpr/gateway/user_details.rb
koetsier: Yes, but it takes a month before the inactive users are actually deleted right? So between 11 and 12 months of inactivity do users not get emailed every day?

koetsier (changes requested):

lib/gdpr/gateway/user_details.rb
```diff
@@ -79,12 +75,47 @@ def notify_inactive_users
...
+  def find_inactive_users(days)
+    WifiUser::User.where(
+      Sequel.lit("
+      (last_login < DATE_SUB(NOW(), INTERVAL ? DAY)
+      OR (last_login IS NULL AND created_at < DATE_SUB(NOW(), INTERVAL ? DAY)))
+      AND username != 'HEALTH'", days, days),
+    ).all
+  end
+
+  def find_users_inactive_for_11_months(days)
```
koetsier, line 105: this name is a bit confusing. It looks for users that are inactive for exactly a to be defined number of days, not necessarily 11 months right?

koetsier (approved):
one small comment

PR #820, opened 2024-08-01
exclude images from the email extractor

iram-shehzadi:
### What
Email extractor picking up images

### Why
Remove images from the email extractor


koetsier (approved):
👍

PR #821, opened 2024-08-01
decouple running state from AWS environment

koetsier:
### What
What
The ruby application running state (production, development, test) needs to be decoupled from the AWS
 environment it’s running in.

Why
Leaving things as they are means that spinning new GovWifi environments from scratch take extra time, and this delays the ability to bring the service back online in the event of a serious incident.

- Introduce a NotifyTemplate class to index templates from Notify
- Remove numerous pointless indirections
- Ensure testing sending Notify messages is consistent throughout the app

Link to Jira card (if applicable): 
https://technologyprogramme.atlassian.net/browse/GW-1742


szd55gds (approved):
LGTM, checked for obvious typo's.

PR #841, opened 2024-09-16
Notify error should raise an error to be captured by Sentry

koetsier:
The `ensure` statement meant the error was never raised and sent to Sentry


szd55gds (approved):
LGTM

PR #843, opened 2024-09-18
Introduce a custom error class to capture application specific errors

koetsier:
So that more errors are being sent to Sentry if something is to go wrong. Any expected errors caused by malformed user content being sent to the User signup API are caught using the UserSignupError class.

All other unspecified errors are now sent to Sentry to warn us about a potential issue in the code / enviromnent

### What
Describe the change

### Why
Describe why the change was necessary

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -39,7 +39,7 @@ class App < Sinatra::Base
   end
 
   post "/user-signup/email-notification" do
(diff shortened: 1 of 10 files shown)


leonaAtkins (approved):
LGTM

PR #850, opened 2024-09-23
Fix syntax on the migration

koetsier:
This works - the previous form did not. Also, ensure all records have the follow_up_sent_at attribute set, which ensures none of the current users will be sent a follow-up message


szd55gds (approved):
LGTM, merge when ready

PR #859, opened 2024-10-08
Use support reply id when sending followup emails

koetsier:
The followup email for inactive users should have the address of our support email inbox as a reply-to field so that users can contact us with their questions.

### DIFF
--- a/lib/wifi_user/email_sender.rb
+++ b/lib/wifi_user/email_sender.rb
@@ -63,7 +63,7 @@ class WifiUser::EmailSender
     Services.notify_client.send_email(
       email_address:,
       template_id: Notifications::NotifyTemplates.template(:followup_email),
-      email_reply_to_id: do_not_reply_email_address_id,
+      email_reply_to_id: support_reply_email_address_id,
     )
   end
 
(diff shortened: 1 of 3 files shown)


lib/wifi_user/email_sender.rb
```diff
@@ -100,4 +100,8 @@ def self.send_active_users_signup_survey(user)
   def self.do_not_reply_email_address_id
     ENV["NOTIFY_DO_NOT_REPLY"] || "0d22d71f-afa3-4c72-8cd4-7716678dbd43"
   end
+
+  def self.support_reply_email_address_id
+    ENV["NOTIFY_SUPPORT_REPLY"] || "5619b95b-2f93-4a70-be64-2f2568f8876f"
```
leonaAtkins, line 105: oh I don't think this UUID should be hidden here, could it be a constant or somehow easy to find? I'm sure it'll be different on each env?

koetsier: It's just a default so that if the variable isn't set we use that (which so happens to be the production reply-to-id) In practice we shouldn't really need it

leonaAtkins (approved):

PR #897, opened 2024-12-19
Ensure fewer notify errors  get sent to Sentry

koetsier:
BadRequestErrors (400) are non-fatal errors such as the phone number not being formatted correctly. This is a more
specific error than RequestErrors and these should not end up in Sentry, but should be logged.

https://technologyprogramme.atlassian.net/browse/GW-2059


mel-am (approved):
lgtm

PR #937, opened 2025-03-26
GW-1821 Backup notify templates

ripan:
### What

Create a  rake task `backup_notify_templates` to backup notify templates to an S3 bucket

### Why
To back up and recover our Notify templates, in case there is a DR scenario. 

> [!IMPORTANT]  
> Currently there is no way to create notify templates Programmatically. In a DR scenario, one has to create the templates manually on notifications service https://www.notifications.service.gov.uk/services/:id/templates 

[GW-1821](https://technologyprogramme.atlassian.net/browse/GW-1821)


spec/lib/common/gateway/s3_object_uploader_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      expect(mock_s3_client).to receive(:create_bucket).with(
+        bucket: bucket,
+        create_bucket_configuration: { location_constraint: region },
+      )
+
+      subject.send(:create_bucket)
+    end
+  end
+
+  describe "#upload_file" do
```
leonaAtkins, line 51: Do you have a test for if the upload fails?  If it fails will it delete the file to be backed up, or will it exit?

ripan: @leonaAtkins File exists even if upload fails. Added a spec for error scenario

leonaAtkins (approved):

lib/common/gateway/s3_object_uploader.rb
```diff
@@ -0,0 +1,53 @@
...
+        logger.info "Bucket '#{bucket}' created"
+      rescue Aws::S3::Errors::BucketAlreadyOwnedByYou => e
+        logger.warn "Bucket '#{bucket}' already exists: #{e.message}"
+      end
+
+      def upload_file(file_path, s3_key)
+        s3.put_object(
+          bucket: bucket,
+          key: s3_key,
+          body: File.open(file_path, "rb"),
```
koetsier, line 47: I think body can also be a string so you don't have to save the templates to a temp file

spec/lib/common/gateway/s3_object_uploader_spec.rb
```diff
@@ -0,0 +1,92 @@
+describe Common::Gateway::S3ObjectUploader do
```
koetsier, line 1: I feel some of these tests use a bit too much mocking; they end up testing if each line of the implementation is called in sequence and so have too much knowledge of the implementation.

I think a better way is to use AWS' built-in mocking system(https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/stubbing.html) . 

We already stub out put_object and get_object in the ```spec_helper``` file so that whenever you use put_object, you can retrieve that same object using get_object.

koetsier (changes requested):
some comments. let me know what you think

spec/lib/common/gateway/s3_object_uploader_spec.rb
ripan: @koetsier Thanks for the suggestion. It simplified my specs allot

docker-compose.yml
```diff
@@ -23,6 +23,7 @@ services:
       RACK_ENV: development
       S3_SIGNUP_ALLOWLIST_OBJECT_KEY: 'somefile'
       S3_SIGNUP_ALLOWLIST_BUCKET: 'somebucket'
+      S3_NOTIFICATION_TEMPLATES_BUCKET: 'somebucket'
```
koetsier, line 26: might want to call this bucket something different so it can't clash with S3_SIGNUP_ALLOWLIST_BUCKET

lib/common/gateway/s3_object_uploader.rb
```diff
@@ -0,0 +1,51 @@
...
+      def ensure_bucket_exists
+        return if bucket_exists?
+
+        create_bucket
+      end
+
+      def bucket_exists?
+        s3.head_bucket(bucket: bucket)
+        true
+      rescue Aws::S3::Errors::NotFound, Aws::S3::Errors::Forbidden
```
koetsier, line 38: Aws::S3::Errors::Forbidden should be handled differently because in this case, creating a bucket would probably fail

lib/notifications/export_templates.rb
```diff
@@ -0,0 +1,36 @@
...
+      body
+      letter_contact_block
+      version
+      created_at
+      updated_at
+      created_by
+    ].freeze
+
+    def self.execute
+      body = StringIO.new(templates.to_json)
```
koetsier, line 21: I don't think you need StringIO, you can just use a string in the upload method. The templates are already in memory and don't need to be streamed.

koetsier (changes requested):

docker-compose.yml
ripan: @koetsier  The bucket name would be `"${var.app_env}_notification_templates"`. Or is there any convention i need to follow while naming a bucket?
https://github.com/GovWifi/govwifi-terraform/pull/939/files#diff-0781b352b0672b84a7884f30765b95ee45fb70036c9c9c7715096d8e8030ad92R135

lib/common/gateway/s3_object_uploader.rb
ripan: @koetsier In this case I prefer rescue only `NotFound `. and let 'Forbidden' throws an error. Any suggestion to handle it differently?

.gitignore
```diff
@@ -1,3 +1,4 @@
 # SimpleCov output
 coverage
 .idea
```
koetsier, line 3: don't need this

koetsier (approved):
👍

PR #1118, opened 2026-05-05
FIX: Dependabot - gem "net-imap", ">= 0.5.14"

ethancane-digital-co:
### What
Several vulnerabilities in the net-imap gem have been discovered.
It is recommended to update to the latest version of this gem to fix these.

### Why
To prevent several high/moderate attack vectors based around escalation highjacking.

[Dependabot](https://github.com/GovWifi/govwifi-user-signup-api/security/dependabot)


ethancane-digital-co:
This fix has already been applied.

